Issue #1, opened 2017-01-24
Size of input should be a natural number


faheel:
Sorting programs may crash if `size` of input values is negative.
Modify the function [`getInputSize`](../../blob/master/Sorting/SortingUtils.h#L18-L21) to check that `size >= 0`. In case it's not, then display:
>     Invalid input size! Try again.
and prompt to input `size` again (through a recursive call to `getInputSize`).

Also add a condition, that if `size` is 0, the program should display:
>     Nothing to sort here.
and exit.

faheel:
Issue fixed in #2.

PR #2, opened 2017-01-25
Fix issue #1: Size of input should be a natural number. 

ampmangu:
It would be better to do this throwing an exception to the main method and catching it, but in this way we avoid having to modify all programs.

### DIFF
--- a/Sorting/SortingUtils.h
+++ b/Sorting/SortingUtils.h
@@ -9,8 +9,10 @@
 
 #ifndef VECTOR_H
 #include <vector>
-#endif
+#include <cstdlib>
 
+#endif
+#define EXIT_CODE_ERROR -1
 using namespace std;
 
 // Input utils:
@@ -18,6 +20,13 @@ using namespace std;
 void getInputSize(size_t& size) {
     cout << "Enter the input size : ";
     cin >> size;
+    if(size <0) {
+        cout << "Invalid input size! Try again.\n";
+        getInputSize(size);
+    }else if(size == 0) {
+        cout << "There is nothing to sort here\n";
+        exit(EXIT_CODE_ERROR );
+    }
 }
 
 void getInputValues(vector<int>& values, const size_t& size) {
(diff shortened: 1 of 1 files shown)


Sorting/SortingUtils.h
```diff
@@ -9,15 +9,24 @@
...
+#endif
+#define EXIT_CODE_ERROR -1
 using namespace std;
 
 // Input utils:
 
 void getInputSize(size_t& size) {
     cout << "Enter the input size : ";
     cin >> size;
+    if(size <0) {
```
faheel: Since `size` is of type `size_t`, which is an unsigned integer, it will never _actually_ be negative, and this condition will always be false!
Eg. if the value entered for `size` is `-5`, in binary it will be (assuming an `int` is 4 bytes long) `10000000000000000000000000000101` which for a `size_t` in decimal is `2147483653`.

It would therefore be right to change this line to:
```C++
    if ((int) size < 0) {
```

What are your thoughts on this?

Sorting/SortingUtils.h
```diff
@@ -9,15 +9,24 @@
 
 #ifndef VECTOR_H
 #include <vector>
-#endif
+#include <cstdlib>
 
+#endif
+#define EXIT_CODE_ERROR -1
```
faheel: Give it a more meaningful name such as `EXIT_SORT_NOT_REQUIRED`.

Also, since `-1` is the code used mostly for errors, give it a value of, say, `2`. (Since it is not _really_ an error if there are 0 elements to sort, right?)

faheel (changes requested):

Sorting/SortingUtils.h
ampmangu: Hi! First, let me thank you for the review, it's the first time I do a pull request, so I appreciate the feedback. I love the idea with your project, and I forked it because I wanted to improve my basics knowledge on C++ (more basic than I remembered as I see :( ) but, thank you, I will fix this, and I'm happy to help in other issues.

faheel:
Closes #1

faheel:
Thanks @Mangu93!

PR #3, opened 2017-01-26
KADANE'S ALGORITHM

himanshu14263:
This algo helps to find the largest sum continuous array.


himanshu14263:
making my first contribution

find the sum of the largest sum continuous array.cpp
```diff
@@ -0,0 +1,85 @@
+// PROBLEM :: find the sum of the largest sum continuous array.
+
+#include <bits/stdc++.h>
```
faheel: Please include only those header files that are required for this program.

find the sum of the largest sum continuous array.cpp
```diff
@@ -0,0 +1,85 @@
+// PROBLEM :: find the sum of the largest sum continuous array.
+
+#include <bits/stdc++.h>
+using namespace std;
+
+// METHOD 1 : USING KADANE'S ALGORITHM
+// DOESN'T WORK IF ALL THE ELEMENTS ARE NEGATIVE
```
faheel: Since this method fails when all elements are negative, while method 2 doesn't, it's better to have just method 2 and remove method 1. After all, method 2 is also Kadane's algorithm.

find the sum of the largest sum continuous array.cpp
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	cout<<max_so_far<<"\n";
+	for(int i=s;i<=e;i++)
+		cout<<a[i]<<" ";
+}
+int main()
+{
+
+	int arr[]={-2,-3,4,1,-2,-1,5,-3};
```
faheel: Take input from the user instead of hard-coding the elements.
Also, I suggest that you define a `vector<int>` instead of a C-style array.

find the sum of the largest sum continuous array.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+	int n = sizeof(arr)/sizeof(arr[0]);
```
faheel: Ask the user for the number of elements and input those many in the vector. This line wouldn't be required then.

find the sum of the largest sum continuous array.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+	cout<<"METHOD 1 :: \n";
+	kadaneAlgo(arr,n);
+
+	cout<<"\n\n";
+
+	cout<<"METHOD 2 :: \n";
+	maxSumSubArray(arr,n);
```
faheel: Also print the starting and ending positions (index + 1) of the sequence that gives the largest sum.

faheel (changes requested):
- Rename the file as `Kadane.cpp` and save it in a folder named `Largest Sum Contiguous Sub-sequence`.

- Take input from the user. Eg.
  ```
  Enter the number of elements : 10
  Enter 10 elements :
  3 6 -5 2 4 -7 8 -9 5 2
  ```
- Use proper variable names such as `values` and `size`, instead of `a` and `n`.
- The output of the program should be like:
  ```
  Maximum contiguous sum is 11.
  Sequence starts at position 1 and ends at position 7.
  ```

largest continuous subset/kadane.cpp
```diff
@@ -0,0 +1,56 @@
+// PROBLEM :: find the sum of the largest sum continuous array using KADANE'S ALGORITHM
+
+
+#include <iostream>
+#include <vector>
+using namespace std;
+
+
+void maxSumSubArray(vector<int>values)
```
faheel: Take `values` as a reference to avoid creating a copy. Also, since the elements in `values` aren't modified, use the `const` modifier.

So you can change `vector<int>values` to `const vector<int>& values`.

largest continuous subset/kadane.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+	int max_so_far=values[0];
+	int curr_max=values[0];
+
+	int s=0,e=0;
```
faheel: Change `s` and `e` to `start` and `end` respectively.

largest continuous subset/kadane.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+	int n=values.size();
+
+	for(int i=1;i<n;i++)
```
faheel: You do not need to take an extra variable `n` to store the size of `values`.
You can instead do:
```C++
    for (int i = 1; i < values.size(); i++) {
        ... 
```
This wouldn't be a performance issue, and would improve code readability too. You may read [this StackOverflow answer](http://stackoverflow.com/a/3901666/5245129) for more on this topic.

largest continuous subset/kadane.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+	cout<<"the max contiguous sum is :: " << max_so_far<<"\n";
+	cout<<"the sequence starts from "<<s<<" and ends at "<<e<<"\n";
+	cout<<"and the values are :: ";
+	for(int i=s;i<=e;i++)
+		cout<<values[i]<<" ";
+}
+int main()
+{
+	int n;
```
faheel: Change `n` to `size`, and change its type to [`size_t`](http://en.cppreference.com/w/cpp/types/size_t).

largest continuous subset/kadane.cpp
```diff
@@ -0,0 +1,56 @@
...
+	vector<int>values;
+	cout<<"how many values you want to enter in the array \n";
+	cin>>n;
+	cout<<"enter the values in the array\n";
+	int x;
+	for(int i=0;i<n;i++)
+	{
+		cin>>x;
+		values.push_back(x);
+	}
```
faheel: Declare the vector as:
```C++
    vector<int> values(size);
```
after taking the input for `size`.
Then use a [range-based for-loop](http://en.cppreference.com/w/cpp/language/range-for) to input values of its elements like so:
```C++
    for (int& val : values) {
        cin >> val;
    }
```

faheel (changes requested):
Most of it looks good! However, there are a few changes that need to be made. After that, it'll be good to merge!

largest continuous subset/kadane.cpp
himanshu14263: Why it is necessary? Just curious

faheel (approved):

faheel:
Looks good @lelouch-lampourge. Thank you for your time and contribution! :smiley:

himanshu14263:
could you please guide me about how i can do open source contrubution to a community

faheel:
By a "community", do you mean "web development", "data science" communities etc.
If so, then look for the most popular/useful open-source projects (libraries, frameworks etc.) in those areas and contribute to those.

largest continuous subset/kadane.cpp
faheel: Read [this](http://stackoverflow.com/a/1951547/5245129).

himanshu14263:
Ok i'll look into it, thank you !

Issue #5, opened 2017-01-29
Add implementation of Disjoint Set data structure


torquecoder:

faheel:
Addressed in #7

PR #7, opened 2017-01-30
Implemented disjoint set data structure

quantum-pie:
Basic disjoint set implementation with path compression and union-by-rank.

Addresses issue #5

### DIFF
--- a/DataStructures/disjoint.cpp
+++ b/DataStructures/disjoint.cpp
@@ -1,0 +1,40 @@
+#include "disjoint.h"
+
+disjoint_set_t::disjoint_set_t(size_t nodes) : parent_(nodes), rank_(nodes, 0)
+{
+    /* initially all elements are singletons i.e. they are parents of themselves */
+    for(size_t i = 0; i < nodes; ++i) {
+        parent_[i] = i;
+    }
+}
+
+int disjoint_set_t::find_set(int x)
+{
+    /* if x is set representative, we are done */
+    if(parent_[x] == x) {
+        return x;
+    }
+    /* travel recursively to the representative node
+       also path compression is done: representative becomes parent of all nodes on the path
+    */
+    return parent_[x] = find_set(parent_[x]);
+}
+
+void disjoint_set_t::union_sets(int x, int y)
+{
+    /* find representatives of inputs (tree roots) */
+    x = find_set(x);
+    y = find_set(y);
+
+    /* decide to connect one root to another based on representatives rank */
+    if(rank_[x] < rank_[y]) {
+        parent_[x] = y;
+    } else {
+        parent_[y] = x;
+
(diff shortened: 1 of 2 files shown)


DataStructures/disjoint.cpp
```diff
@@ -0,0 +1,40 @@
...
+    x = find_set(x);
+    y = find_set(y);
+
+    /* decide to connect one root to another based on representatives rank */
+    if(rank_[x] < rank_[y]) {
+        parent_[x] = y;
+    } else {
+        parent_[y] = x;
+
+        /* inscrease final rank if ranks are identical */
```
faheel, line 35: Typo in `increase`.

DataStructures/disjoint.cpp
```diff
@@ -0,0 +1,40 @@
...
+int disjoint_set_t::find_set(int x)
+{
+    /* if x is set representative, we are done */
+    if(parent_[x] == x) {
+        return x;
+    }
+    /* travel recursively to the representative node
+       also path compression is done: representative becomes parent of all nodes on the path
+    */
+    return parent_[x] = find_set(parent_[x]);
```
faheel, line 20:
```C++
parent_[x] = find_set(parent_[x]);
return parent_[x];
```
would be better.

DataStructures/disjoint.h
```diff
@@ -0,0 +1,20 @@
+#ifndef DISJOINT_H
+#define DISJOINT_H
+
+#include <cstdlib>
+#include <vector>
+
+class disjoint_set_t
```
faheel, line 7: Why `disjoint_set_t`? Why not just `disjoint_set`, or better yet, `DisjointSet`.

([Type Names - Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Type_Names))

DataStructures/disjoint.h
```diff
@@ -0,0 +1,20 @@
...
+{
+public:
+    disjoint_set_t(size_t nodes);
+
+    int find_set(int x);
```
faheel, line 12: `find_set` is kind of a "misleading" name for this function, I think. Here are a few suggestions:
- `root`
- `parent`
- `get_root` or `find_root`
- `get_parent` or `find_parent`

faheel (changes requested):
- Rename the files to `DisjointSet.h` and `DisjointSet.cpp`
- Single-line comments using `//` please!

faheel:
@quantum-pie There are a few changes that I've mentioned in the review. Please take a look so that we can merge this as soon as possible.

faheel:
@quantum-pie There are a few changes that I've mentioned in the review. Please take a look so that we can merge this as soon as possible.

In case you won't be able to do these changes, let me know so that I can do these myself.

PR #9, opened 2017-01-31
Added required description comment in binary_search.cpp

varshika03:
Review it so that I can add more.

### DIFF
--- a/Searching/BinarySearch.cpp
+++ b/Searching/BinarySearch.cpp
@@ -2,7 +2,7 @@
     Binary search: A search algorithm that finds the position of a target value
     within a sorted array
     Time complexity: O(logn), where n is the number of elements in the array.
-    Space complexity: O(n), where n is the number of elements in the array.
+    Space complexity: O(1), where n is the number of elements in the array.
 */
 
 #include <iostream>
--- a/Searching/BinarySearch.cpp
+++ b/Searching/BinarySearch.cpp
@@ -1,13 +1,8 @@
 /*
-    <Binary Search>:
-    <A search algorithm that finds the position of a target value
-    within a sorted array>
-
-    <Time complexity>:
-    O(logn), where <n is the number of elements in the array>    
-    
-    <Space complexity>:
-    O(n), where <n is the number of elements in the array>
+    Binary search: A search algorithm that finds the position of a target value
+    within a sorted array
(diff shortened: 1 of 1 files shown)


quantum-pie:
I think space complexity is O(1) because you don't need additional memory

Searching/BinarySearch.cpp
```diff
@@ -1,7 +1,13 @@
 /*
-    Binary search:
-    A search algorithm that finds the position of a target value
-    within a sorted array
+    <Binary Search>:
+    <A search algorithm that finds the position of a target value
+    within a sorted array>
+
+    <Time complexity>:
+    O(logn), where <n is the number of elements in the array>
```
faheel: I suggest writing `O(logn)` as either `O(logN)`, `O(log(n))` or `O(log(N))` for more clarity.

`O(log(N))` would be the best way IMO.

faheel (changes requested):
Remove all the angle brackets (`<` & `>`). They are used in the format just for showing that the enclosing text is a description of the text, and not the actual text.

Searching/BinarySearch.cpp
```diff
@@ -1,7 +1,13 @@
...
-    within a sorted array
+    <Binary Search>:
+    <A search algorithm that finds the position of a target value
+    within a sorted array>
+
+    <Time complexity>:
+    O(logn), where <n is the number of elements in the array>    
+    
+    <Space complexity>:
+    O(n), where <n is the number of elements in the array>
```
faheel: It's actually `O(1)` as only a constant amount of **extra memory** is required.

faheel: It's actually `O(1)` as only a constant amount of **extra memory** is required.

faheel (changes requested):

faheel (changes requested):
Newlines please! As in this:
```C++
/*
    <Name of algorithm>:
    <Brief description>

    Time complexity:
    O(...), where <description of variable(s)>    
    
    Space complexity:
    O(...), where <description of variable(s)>
*/
```

And again, I suggest writing `O(logn)` as either `O(logN)`, `O(log(n))` or `O(log(N))` for more clarity.
I think `O(log(N))` would be the best way.

faheel:
Closing this since there hasn't been any activity for a long time.

PR #12, opened 2017-02-08
Fibonacci now runs with O(1) space complexity

ilayaraja97:
Fibonacci runs in space complexity of O(1) now and can take bigger inputs to the order of 10^5. Tested for such inputs too. `assert()` used to check bounds.

### DIFF
--- a/NumberTheory/Fibonacci.cpp
+++ b/NumberTheory/Fibonacci.cpp
@@ -5,10 +5,10 @@
     after the first two is the sum of the two preceding ones.)
 
     Time complexity:
-    O(n), where n is the value for which Fibonacci number must be generated
+    O(N), where N is the term of the Fibonacci sequence to calculate
 
     Space complexity:
-    O(1), where n is the value for which Fibonacci number must be generated
+    O(1)
 */
 
 #include <iostream>
@@ -17,10 +17,6 @@ using namespace std;
 
 typedef unsigned long long int ULL;
 
-/* fibonacci recursive
- * time complexity O(n) space complexity O(1)
- * tested for input size < 10^5
- */
 const int MAX_N = 93;   // fibonacci(94) goes beyond the range of ULL
 ULL fibonacci(const ULL n,ULL &previous)
 {
--- a/NumberTheory/Fibonacci.cpp
+++ b/NumberTheory/Fibonacci.cpp
@@ -6,30 +6,44 @@
 */
 
 #include <iostream>
-#include <vector>
+#include <assert.h>
 
 using namespace std;
 
 typedef unsigned long long int ULL;
 
-const int MAX_N = 93;   // fibonacci(94) goes beyond the range of ULL
-
-ULL fibonacci(const int &N, vector<ULL> &fiboMemo) {
-    if (N == 0 or N == 1)
-        fiboMemo[N] = N;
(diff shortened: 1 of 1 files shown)


faheel:
I think you're mistaken by what "input" means here.

The input, `N`, is for which term of the Fibonacci sequence is to be calculated. And since `Fibonacci(94)` (which is 1.97\*10^19 approx.) is the first term of the sequence that goes beyond the range of `unsigned long long int` (which is 1.84\*10^19 approx.), 93 is the max limit for `N`.

As for the reduced space complexity, well done!

faheel (changes requested):
I think you're mistaken by what "input" means here.

The input, `N`, is for which term of the Fibonacci sequence is to be calculated. And since `Fibonacci(94)` (which is 1.97\*10^19 approx.) is the first term of the sequence that goes beyond the range of `unsigned long long int` (which is 1.84\*10^19 approx.), 93 is the max limit for `N`.

So you can leave `getN()` the way it was, and remove any `assert()` as it wouldn't be required, given the base conditions for when `N` is 0 or 1.

As for the reduced space complexity, well done!

ilayaraja97:
* removed use of `assert()` 
 * left getN() the way it was

I'm quite surprised about how I missed the comment saying values above 93 overflows capacity of `unsigned long long`.
Any more issues? It's my first contribution in a public repo.

NumberTheory/Fibonacci.cpp
```diff
@@ -3,6 +3,12 @@
     Find the N-th Fibonacci number, given the value of N.
     (The Fibonacci sequence is characterized by the fact that every number
     after the first two is the sum of the two preceding ones.)
+
+    Time complexity:
+    O(n), where n is the value for which Fibonacci number must be generated
+
+    Space complexity:
+    O(1), where n is the value for which Fibonacci number must be generated
```
faheel: Just `O(1)` is enough.
No need of `, where n ...` here.

NumberTheory/Fibonacci.cpp
```diff
@@ -3,6 +3,12 @@
     Find the N-th Fibonacci number, given the value of N.
     (The Fibonacci sequence is characterized by the fact that every number
     after the first two is the sum of the two preceding ones.)
+
+    Time complexity:
+    O(n), where n is the value for which Fibonacci number must be generated
```
faheel: Use `N` instead of `n`.

Also, I think the following:
> where N is the term of the Fibonacci sequence to calculate

would better describe what `N` is. What do you think?

faheel (changes requested):

faheel (approved):

faheel (changes requested):
Another small change and then you'll be good to merge!

NumberTheory/Fibonacci.cpp
```diff
@@ -3,25 +3,44 @@
...
-#include <vector>
 
 using namespace std;
 
 typedef unsigned long long int ULL;
 
+/* fibonacci recursive
+ * time complexity O(n) space complexity O(1)
+ * tested for input size < 10^5
+ */
```
faheel: This multi-line comment can be removed.

ilayaraja97:
Please review it again.

faheel (approved):

PR #24, opened 2017-02-26
Added Recursive MaxSubarray

ScottGBennett:
Kadane's algorithm is more efficient, but it may be helpful to have a recursive implementation to contrast with.

### DIFF
--- a/MaximumSubarray/RecursiveSubarray.cpp
+++ b/MaximumSubarray/RecursiveSubarray.cpp
@@ -1,0 +1,89 @@
+// This is a recursive algorithm to solve the Maximum Subarray Sum Problem
+// As a recursive, divide-and-conqure algorithm, this runs in O(n log n)
+// Thus, Kadane's algorithm (which runs in O(n)) is more efficient and a better choice.
+// This is included as an example of a recursive solution.
+
+#include <stdio.h>
+
+// helper function - returns the maximum of 2 integers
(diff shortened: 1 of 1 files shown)


MaximumSubarray/RecursiveSubarray.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+int main()
+{
+   //the contiguous subarray with the largest sum is 4, -1, 2, 1, with sum 6.
+   //see     https://en.wikipedia.org/wiki/Maximum_subarray_problem
+   //for answer verification
+   int theArray[] = {-2,1,-3,4,-1,2,1,-5,4};
+   int n = sizeof(theArray)/sizeof(theArray[0]);
+   int maxSum = maxSubArraySum(theArray, 0, n-1);
+   printf("Maximum sum of continuous array is %d\n", maxSum);
```
faheel, line 87: It would be better to take input from the user, rather than hard-coding the values. You can take the input in the same format as in [Kadane.cpp](../../blob/master/MaximumSubarray/Kadane.cpp)

faheel (changes requested):

faheel:
Yes, it would be good to add this as a comparison to Kadane's algorithm.

faheel:
Closing this PR since there hasn't been any update since quite a long time.

PR #33, opened 2017-03-20
changed Relational operators

arielyyz:
Fixed operators logic and const correctness.
Removed code duplication.
Added tests for operators.

### DIFF
--- a/DataStructures/BigInteger/BigInt.cpp
+++ b/DataStructures/BigInteger/BigInt.cpp
@@ -96,6 +96,15 @@ void BigInt::operator=(const std::string num) {
     --------------------
 */
 
+BigInt operator+(const BigInt& num) const {
+    BigInt result(max(num, *this));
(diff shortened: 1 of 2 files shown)


DataStructures/BigInteger/BigInt.cpp
```diff
@@ -96,6 +96,15 @@ void BigInt::operator=(const std::string num) {
     --------------------
 */
 
+BigInt operator+(const BigInt& num) const {
+    BigInt result(max(num, *this));
+    BigInt added(min(num, *this));
+    int carry = 0;
```
faheel: Remove this from the commit since this is incomplete and is also not related with relational operators (which your PR is about).

faheel (changes requested):
- Use `and` instead of `&&` and `or` instead of `||`

faheel:
Thanks for your help @arielyyz, and congratulations on your first open source contribution!

image

PR #38, opened 2017-03-23
Add Binary Search Tree

varunjm:
- Adds binary search tree under a new 'Trees' folder.
- Includes insert, remove, Inorder traversal, Preorder traversal, Postorder traversal. Both iterative and recursive.
- Has the time and space complexities where ever applicable

### DIFF
--- a/Trees/BinarySearchTree.cpp
+++ b/Trees/BinarySearchTree.cpp
@@ -1,0 +1,416 @@
+/*
+    Author:
+    Varun Jayathirtha (www.github.com/varunjm)
+
+    Binary Search Tree:
+    A graph data structure where each node has a unidirectional path to at most
+    two node, none of which are itself or its predecessors, i.e the graph is
+    acyclic. The first node is called the root node. Another important feature
+    of this tree is that each node has values lesser than or equal to itself in
+    the left subtree and values greater than or equal to itself in the right
+    subtree.
+*/
+
+#include "BinarySearchTree.hpp"
+#include <iostream>
+#include <stack>
+
+using namespace std;
+
+Node* get_node() {
+    Node* newNode = new Node();
+    newNode->left_child = newNode->right_child = nullptr;
+
+    return newNode;
+}
+
+BinarySearchTree::BinarySearchTree() {
+    root = get_node();
+    root->value = INF;
+}
+
+/*
+    Insert
+    This traverses the tree to find the right spot to insert the newly given
+    value. At each node, it checks if the node has a value lesser than or equal
+    to itself, if so it goes to its left subtree. Otherwise it goes to the right
+    subtree. When it meets an end point i.e. NULL it inserts the new Node along
+    with the new value
+
+    Time complexity:
+    Average case : O(log n), where n is the number of nodes in the tree
+    Worst case : O(n)
+
+    Space complexity:
+    O(1)
+*/
+
+bool BinarySearchTree::insert(int value) {
+    // If first insertion, replace the value of the already created node
+    if(root->value == INF) {
+        root->value = value;
+        return true;
+    }
+
+    Node* current = root;
+    Node* newNode = get_node();
+
+    // Failed to allocate memory to a new node
+    if (newNode == nullptr) {
+        return false;
+    }
+
+    while (true) {
+        if (value <= current->value) {
+            if (current->left_child == nullptr) {
+                current->left_child = newNode;
+                current->left_child->value = value;
+                return true;
+            } else {
+                current = current->left_child;
+            }
+        } else {
+            if (current->right_child == nullptr) {
+                current->right_child = newNode;
+                current->right_child->value = value;
+                return true;
+            } else {
+                current = current->right_child;
+            }
+        }
+    }
+}
+
+/*
+    Remove
+    This traverses the tree to find the given value. At each node, it checks if
+    the node has a value equal to the given value, if so it finds the inorder
+    successor of the node, and replaces the value to be deleted with that number.
+    Otherwise, it searches in the left subtree if the value is smaller than
+    node value and it searches the right subtree if the value is larger.
+
+    Time complexity:
+    Average case : O(log n), where n is the number of nodes in the tree
+    Worst case : O(n)
+*/
+
+bool BinarySearchTree::remove(int value) {
+    // If tree is empty remove is unsucessful
+    if (root->value == INF) {
+        return false;
+    }
+
+    Node* current = root;
+    Node* parent = nullptr;
+    
+    while (true) {
+        if (value == current->value) {
+            remove_current_node(current, parent);
+        } else {
+            parent = current;
+
+            if (current->value < value) {  // Search in the right subtree
+                current = current->right_child;
+            } else {                    // Search in the left subtree
+                current = current->left_child;
+            }
+
+            if (current == nullptr) {
+                return false;
+            }
+        }
+    }
+}
+
+void BinarySearchTree::remove_current_node(Node* current, Node* parent) {
+    Node* toBeDeleted;
+    Node* successor;
+    Node* successorParent = nullptr;
+
+    // If current node has no children
+    if (current->right_child == nullptr and current->left_child == nullptr) {
+        if(parent == nullptr) {
+            current->value = INF;
+        } else {
+            if(parent->left_child == current)
+                parent->left_child = nullptr;
+            else
+                parent->right_child = nullptr;
+
+            delete current;
+        }
+    } else if (current->right_child == nullptr) {
+        toBeDeleted = current->left_child;
+        current->value = current->left_child->value;
+        current->right_child = current->left_child->right_child;
+        current->left_child = current->left_child->left_child;
+        delete toBeDeleted;
+    } else {
+        successor = current->right_child;
+        successorParent = nullptr;
+
+        while(successor->left_child != nullptr) {
+            successorParent = successor;
+            successor = successor->left_child;
+        }
+
+        if (successorParent == nullptr) {
+            current->right_child = successor->right_child;
+        } else {
+            successorParent->left_child = successor->right_child;
+        }
+        current->value = successor->value;
+        delete successor;
+    }
+}
+
+/*
+    Search
+    This traverses the tree to find the presence of number in the tree. At each 
+    node, it checks if the node has a value lesser than itself, if so it goes 
+    to its left subtree. If it is greater it goes to the right subtree. When it 
+    node with the matching value or reaches the end of the tree it returns either 
+    true or false appropriately.
+
+    Time complexity:
+    Average case : O(log n), where n is the number of nodes in the tree
+    Worst case : O(n)
+
+    Space complexity:
+    O(1)
+*/
+
+bool BinarySearchTree::search(int value) {
+    Node* current = root;
+
+    while (true) {
+        if (value < current->value) {
+            if (current->left_child == nullptr) {
+                return false;
+            } else {
+                current = current->left_child;
+            }
+        } else if(value > current->value) {
+            if (current->right_child == nullptr) {
+                return false;
+            } else {
+                current = current->right_child;
+            }
+        } else {
+            return true;
+        }
+    }
+}
+
+/*
+    In order traversal:
+    This Algorithm visits and displays each node in the following order starting
+    with the root node. The Left subtree, followed by Parent, followed by the
+    Right subtree. This traversal gives the values stored in the tree in a
+    sorted order.
+
+    Time complexity:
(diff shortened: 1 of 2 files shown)


Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,341 @@
...
+#include <iostream>
+#include <climits>
+#include <stack>
+
+using namespace std;
+
+typedef struct treeNode {
+    int value;
+    struct treeNode * leftChild;
+    struct treeNode * rightChild;
```
faheel: A node can be defined more simply like:
```C++
struct Node {
    int value;
    Node* left_child;
    Node* right_child;
};
```

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,341 @@
...
+} NODE;
+
+NODE * getNode() {
+    NODE * newNode = new NODE();
+    newNode->leftChild = newNode->rightChild = NULL;
```
faheel: Use [`nullptr`](http://en.cppreference.com/w/cpp/language/nullptr) instead of `NULL` for a null pointer.

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+    return newNode;
+}
+
+NODE * createTree() {
+    NODE * newNode = getNode();
+    newNode->value = INT_MAX;
+
```
faheel: Also initialise the left and right children of `newNode` as `nullptr`

faheel (changes requested):
I think it would be better to declare the BST as a class (named `BinarySearchTree`), having:
- Data members:
  - `root` of type `Node*`
- Member functions:
  - a constructor for initialising the `root` node
  - `insert_node()`
  - `remove_node()`
  - `traverse_inorder()`
  - `traverse_preorder()`
  - `traverse_postorder()`

Also, it would be better to separate the declarations into a header file called `BinarySearchTree.hpp` and the definitions for them in `BinarySearchTree.cpp`

Trees/BinarySearchTree.cpp
varunjm: getNode() initializes all new nodes with left and right children as null

varunjm:
@faheel The requested changes have been made.

Trees/BinarySearchTree.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef BINARY_SEARCH_TREE
+#define BINARY_SEARCH_TREE
+
+struct Node {
+    int value;
+    struct Node * left_child;
+    struct Node * right_child;
+};
```
faheel: There's no need to write `struct` when declaring an object of a structure in C++.

Therefore, change the last 2 lines to:
```C++
    Node* left_child;
    Node* right_child;
```

Also note that you should **not declare pointers like `type * name`!** Instead, you should either write `type* name` or `type *name`, though `type* name` is the preferred choice of most programmers. (Don't forget to make this change in other places as well.)

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,348 @@
...
+Node * get_node() {
+    Node * newNode = new Node();
+    newNode->left_child = newNode->right_child = nullptr;
+
+    return newNode;
+}
+
+BinarySearchTree :: BinarySearchTree() {
+    root = get_node();
+    root->value = INT_MAX;
```
faheel: Instead of assigning `INT_MAX` directly, declare a global constant integer named `INF`, set it to `INT_MAX` and use that in place of `INT_MAX`.

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+    Time complexity:
+    Best case : O(log n), where n is the number of nodes in the tree
+    Worst case : O(n)
+
+    Space complexity:
+    O(1)
+*/
+
+bool BinarySearchTree :: insert_node(int value) {
```
faheel: Don't put spaces around `::`

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,348 @@
...
+    // If first insertion, replace the value of the already created node
+    if(root->value == INT_MAX) {
+        root->value = value;
+        return true;
+    } else {
+        Node * iter = root, * newNode = get_node();
```
faheel: Avoid declaring multiple pointers in the same declaration. [See details](https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions).

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,348 @@
...
+    traversal_preorder_helper(root);
+}
+
+void BinarySearchTree :: traversal_preorder_iterative() {
+    if (root->value == INT_MAX) {
+        return;
+    }
+
+    stack<Node *> traversal;
+    Node * iter;
```
faheel: Rename `iter` to `current` or `current_node`.

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,348 @@
...
+bool BinarySearchTree :: remove_node(int value) {
+    // If tree is empty remove is unsucessful
+    if (root->value == INT_MAX) {
+        return false;
+    } else {
+        Node * iter = root, * parent = nullptr, * toBeDeleted, * successor, * successorParent = nullptr;
+
+        while (true) {
+            if (value == iter->value) {
+                if (iter->right_child == nullptr && iter->left_child == nullptr) {
```
faheel: Use `and ` instead of `&&`, and `or` instead of `||`

faheel (changes requested):
- Don't put spaces around `::`
- Avoid declaring multiple pointers in the same declaration
- Use `and ` instead of `&&`, and `or` instead of `||`
- Declare pointers as `type* name;`

faheel:
I just realised that traversal of a data structure isn't a part of. It's just an algorithm that is applied to a data structure to access it's elements.

I therefore think that all the traversal algorithms should be implemented separately in the `Tree/BinarySearchTree/Traversal` folder, and the rest of the code for BST should be added to the `DataStructures/Trees/BinarySearchTree` folder.

@varunjm I would like to know your view on this.

varunjm:
I had thought about it before my PR, but I decided against it.

1. It would lead to a very confusing arrangement for any beginner who is looking to understand the working from this code.
2. In-order, pre-order and post-order traversals are almost always associated with Binary search trees.

Trees/BinarySearchTree.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+Node* get_node();
+
+class BinarySearchTree {
+private:
+    Node* root;
+
+    void traversal_inorder_helper(Node* tree);
+    void traversal_preorder_helper(Node* tree);
+    void traversal_postorder_helper(Node* tree);
```
faheel: I don't think there's a need for these 3 functions. They can be removed (from the _cpp_ file as well)

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,358 @@
...
+    the left subtree and values greater than or equal to itself in the right
+    subtree.
+*/
+
+#include "BinarySearchTree.hpp"
+#include <iostream>
+#include <climits>
+#include <stack>
+#define INF INT_MAX
+
```
faheel: To declare constants, use a `const` "variable", rather than the `#define` directive.
```C++
const int INF = INT_MAX;
```

faheel (changes requested):

Trees/BinarySearchTree.hpp
varunjm: It is a recursive function with the current node being passed as an argument. The function call to the traversal methods using objects makes it unnecessary and not possible to pass the current node as an argument to it. This is needed.

Trees/BinarySearchTree.cpp
varunjm: INT_MAX max is (sort of) a constant declared in the "climits" library. Why should i redeclare it as a constant? The only reason I have this line is to rename the constant to be more meaningful.

varunjm:
@faheel Any views on my comments?

Trees/BinarySearchTree.cpp
faheel: By using a `const` you would be doing the same thing that you are doing with `#define` - giving the constant a better name, but in a more standard way for C++ (`#define` is old C style, modern C has `const` for the same purpose).

Trees/BinarySearchTree.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+public:
+
+    BinarySearchTree();
+    bool insert_node(int);
+    bool remove_node(int);
+    void traversal_inorder_recursive();
```
faheel: The **search** operation is missing, which is the one of the most important operation in a BST. Can you please add it?

Trees/BinarySearchTree.hpp
faheel: If you set the default value of the argument of the recursive functions as `root` (the tree's root), then the helper functions can be removed and the functions can still be called from `main()` (or other external functions) without passing an argument - the way they are called now.

Eg.
```C++
void BinarySearchTree::traverse_inorder_recursive(Node* node = root) {
    if (node == nullptr or node->value == INF)
        return;

    traverse_inorder_recursive(node->left_child);
    std::cout << node->value << '\n';
    traverse_inorder_recursive(node->right_child);
}
```

Trees/BinarySearchTree.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+    void traversal_inorder_helper(Node* tree);
+    void traversal_preorder_helper(Node* tree);
+    void traversal_postorder_helper(Node* tree);
+
+public:
+
+    BinarySearchTree();
+    bool insert_node(int);
+    bool remove_node(int);
```
faheel: I suggest that you rename `insert_node` and `remove_node` to `insert` and `delete` respectively, which are the more conventional names for the operations, followed in many other places including the STL.

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,358 @@
...
+/*
+    Remove
+    This traverses the tree to find the given value. At each node, it checks if
+    the node has a value equal to the given value, if so it finds the inorder
+    successor of the node, and replaces the value to be deleted with that number.
+    Otherwise, it searches in the left subtree if the value is smaller than
+    node value and it searches the right subtree if the value is larger.
+
+    Time complexity:
+    Best case : O(log n), where n is the number of nodes in the tree
```
faheel: _O_(log(N))  is the average case. Best case will be _O_(1), so there's no need to mention it. Mention the average case instead, along with the worst case (for other methods too).

Trees/BinarySearchTree.cpp
```diff
@@ -0,0 +1,358 @@
...
+    Worst case : O(n)
+*/
+
+bool BinarySearchTree::remove_node(int value) {
+    // If tree is empty remove is unsucessful
+    if (root->value == INF) {
+        return false;
```
faheel: Please simplify this function by breaking the different operations that happen in it (like finding the successor etc.) into separate functions, and call those functions here.

faheel (changes requested):

Trees/BinarySearchTree.cpp
varunjm: Okay, I'll change it.

Trees/BinarySearchTree.hpp
varunjm: Okay.

Trees/BinarySearchTree.hpp
varunjm: Sure.

Trees/BinarySearchTree.cpp
varunjm: My bad. Will update this.

Trees/BinarySearchTree.hpp
varunjm: This is disallowed in C++. The best I can do is overload the same function name instead of naming it a 'helper'.

varunjm:
@faheel Any more changes?

faheel:
@varunjm I've troubled you enough :smile:. It's time to merge!

faheel:
Thanks a lot for your contribution and patience! :smiley:

varunjm:
@faheel I dont mind it at all. Great working with you.

Shabalang:
This will also help me. I am open to criticism. Please let me know of any possible improvements I can implement.

PR #31, opened 2017-03-18
BigInt implementation

seanhouser:
Followed instructions provided to begin work on BigInt class

PR #42, opened 2017-04-05
BigInt datatype implementation

seanhouser:
See closed PR #31 for additional history

### DIFF
--- a/DataStructures/BigInteger/BigInt.cpp
+++ b/DataStructures/BigInteger/BigInt.cpp
@@ -4,6 +4,7 @@
 #include "stdafx.h"
 #include <string>
 #include "BigInt.hpp"
+#include <stdexcept>
 
 bool is_valid_number(const std::string& num);
 void notify_invalid_input(const std::string& input);
@@ -38,7 +39,7 @@ int _tmain(int argc, _TCHAR* argv[])
 	std::cout << "The negation of " << unary_test1 << " is: " << -unary_test1 << '\n';
 	*/
 
-	
+	/*
 	// COMPLETE
 	// unary increment tests
 	BigInt unary_test1 = 5;
@@ -54,7 +55,7 @@ int _tmain(int argc, _TCHAR* argv[])
 	std::cout <<  unary_test2 << "--" << " is: ";
 	std::cout << unary_test2-- << " on return from operation, but\n";
 	std::cout << "after return from operation previous BigInt is now: " << unary_test2 << '\n';
-	
+	*/
 	
 	/*
 	// COMPLETE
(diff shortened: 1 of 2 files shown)


DataStructures/BigInteger/BigInt.cpp
```diff
@@ -1,14 +1,261 @@
...
+}
+
+
 /*
     Utility functions
     -----------------
 */
 
 bool is_valid_number(const std::string& num) {
     for (char digit : num)
```
varunjm: Using char& instead will make the code slightly faster.

DataStructures/BigInteger/BigInt.cpp
```diff
@@ -1,14 +1,261 @@
...
-        if (digit < '0' or digit > '9')
-            return false;
+        if (digit < '0' || digit > '9')
+			if (digit != '+' && digit != '-')
+				return false;
```
faheel: `is_valid_number()` checks whether the magnitude of the number (which is a string) is a valid number, i.e. it should only have characters from `0` to `9`. So remove the checks for `+` and `-` because it makes the function work incorrectly.

DataStructures/BigInteger/BigInt.hpp
```diff
@@ -20,53 +20,53 @@ class BigInt {
...
+    bool operator>=(const BigInt&); 
+    bool operator>=(long long);		
+    bool operator<(const BigInt&);	
+    bool operator<(long long);		
+    bool operator<=(const BigInt&);	
+    bool operator<=(long long);		
+    bool operator==(const BigInt&);	
+    bool operator==(long long);		
+    bool operator!=(const BigInt&);	
+    bool operator!=(long long);
```
faheel: Please remove all trailing white-space.

faheel (changes requested):
Please keep the code for relational operators as contributed by @arielyyz [here](https://github.com/faheel/Algos/blob/dev/DataStructures/BigInteger/BigInt.cpp#L133)

faheel:
@seanhouser I have completed the implementation as a separate library: https://github.com/faheel/BigInt

It is divided into separate header files, has unit tests for each operator, and I have used more efficient algorithms for math operators (such as Karatsuba's algorithm for multiplication, long division method for division, etc.). Do check it out!

faheel:
Once BigInt v1 is released, I plan to add its single-include header file to this repo (just like Catch) and use it for implementing algorithms that require large integers (such as some cryptographic algorithms). 

I could have merged this into the `dev` branch but there are still a few errors (such as incorrect return types and parameter types).

So, I'm sorry but I'm closing this PR now. Thanks a lot for your time and contribution!

Issue #8, opened 2017-01-30
Mention time and space complexities in comments


faheel:
For each algorithm, mention its _time complexity_ and _space complexity_ in the "description comment" of its implementation program.

The format for the "description comment" (which is written at the beginning) should be:
```C++
/*
    <Name of algorithm>
    -------------------
    <Brief description>

    Time complexity
    ---------------
    O(...), where <description of variable(s)>    
    
    Space complexity
    ----------------
    O(...), where <description of variable(s)>
*/
```

Following are the algorithms for which time and space complexity hasn't been added yet:
- [x] N-Queens
- [x] Kadane
- [x] Binary search
- [x] Bubble sort
- [x] Counting sort
- [x] Insertion sort
- [x] Merge sort
- [x] Quick sort
- [x] Selection sort

ilayaraja97:
I think its a good idea to put this in the readme or coding guidlines.

faheel:
Yes. Thanks for the suggestion!

Anamika28:
Name of algorithm: Binary Search
    Binary search algorithm is a searching algorithm which uses Divide and conquer technique to search the given element. The prior requirements before applying Binary search is that elements should be present in sorted order.

    Time complexity:
    O(nlogn), where n is the number of input elements    
    
    Space complexity:
    O(n), where n is the number of input elements

faheel:
Time complexity of binary search is _O_(logN) and it's space complexity is _O_(1)

abhishekabk5:
Name of algorithm: Bubble Sort
Bubble sort is a simple sorting algorithm that repeatedly steps through the list to be sorted, compares each pair of adjacent items and swaps them if they are in the wrong order.

Time Complexity:
O(n^2), where n is the number of elements in the list.

Space Complexity:
O(1), temporary space for swapping.

abhishekabk5:
Name of algorithm: Insertion Sort
Insertion sort is a simple sorting algorithm that builds the final sorted array (or list) one item at a time. Less suitable for large lists.

Time Complexity:
O(n^2), where n is the number of elements in the list.

Space Complexity:
O(1), temporary space for swapping.

abhishekabk5:
Name of algorithm: Selection Sort
The selection sort algorithm sorts an array by repeatedly finding the minimum element (considering ascending order) from the unsorted part and putting it at its right place.

Time Complexity:
O(n^2), where n is the number of elements in the list.

Space Complexity:
[...]

PR #46, opened 2017-04-19
Added space and time complexities for Binary search

Anderzun:
Added a bit to #8

### DIFF
--- a/Searching/BinarySearch.cpp
+++ b/Searching/BinarySearch.cpp
@@ -1,7 +1,16 @@
 /*
-    Binary search:
-    A search algorithm that finds the position of a target value
-    within a sorted array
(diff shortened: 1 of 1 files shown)


LouiseBC:
I believe the space complexity should be O(logN) since this is a recursively implemented algorithm.

faheel:
@LouiseBC Space complexity is _O(1)_ because generally the overhead space of the call stack is not considered while calculating the space complexity of recursive functions. Moreover, the implementation of binary search here uses tail recursion which eliminates the stack overhead (if the compiler optimises for it).

faheel:
@AlexanderKunze Thanks for your help, and congratulations on your first PR merge on this project! 🎆

Issue #51, opened 2017-05-05
Add Shell sort 


Hossam-Tarek:
I think Shell sort algorithm will be a good add

PR #52, opened 2017-05-06
Add shell sort algorithm.

Hossam-Tarek:
#51 implementation of shell sort algorithm

### DIFF
--- a/Sorting/ShellSort.cpp
+++ b/Sorting/ShellSort.cpp
@@ -1,0 +1,60 @@
+/*
+	Shell sort
+	-----------
+	in this sorting algorithm we starts by comparing elements that are far apart,
+	then elements that are less far apart and finally comparing adjacent elements.
+	by calculating a gap sequence (gap = n / 2) 
+
+	Time complexity
+	----------------
+	O(N*log(N)^2), where N is the nubmer of elements.
+
+	Space complexity
+	-----------------
+	O(1).
+*/
+
+#include <iostream>
+#include <vector>
+
+using namespace std;
+
+void shell_sort (vector<int> &arr) {
+	int n = arr.size();
+
+	// We calculate the gap sequence in order of |n/2|, |n/4|,....
+	for (int gap = n/2; gap > 0; gap /= 2) {
+		for (int i = gap; i < n; ++i) {
+
+			int temp = arr[i];
+			int j;
+			for (j = i; j >= gap && arr[j - gap] > temp; j -= gap)
+				arr[j] = arr[j - gap];
+
+			arr[j] = temp;
+		}
+	}
+}
+
+// Print the array elements.
+void print(const vector<int> &arr) {
+	for (int i = 0; i < arr.size(); ++i)
+		cout << arr[i] << " ";
+	cout << endl;
+}
+
+int main () {
+	vector<int> arr;
+	int n;
+	cout << "Enter number of elements int the array: ";
+	cin >> n;
+
+	cout << "Enter elements of the array: ";
+	for (int i = 0; i < n; ++i) {
+		int x;
+		cin >> x;
+		arr.push_back(x);
+	}
+	shell_sort(arr);
+	print(arr);
+}
--- a/Sorting/ShellSort.cpp
+++ b/Sorting/ShellSort.cpp
@@ -40,13 +40,6 @@ void shell_sort (vector<int> &values, const bool &show_state) {
     }
 }
 
-// Print the array elements.
-void print(const vector<int> &arr) {
-    for (int i = 0; i < arr.size(); ++i)
-        cout << arr[i] << " ";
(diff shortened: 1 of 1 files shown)


Hossam-Tarek:
@faheel this is the implementation of shell sort algorithm, hope you like this implementation

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,60 @@
+/*
+    Shell sort
+    -----------
+    in this sorting algorithm we starts by comparing elements that are far apart,
+    then elements that are less far apart and finally comparing adjacent elements.
+    by calculating a gap sequence (gap = n / 2)
```
faheel: A few grammatical errors here, like
- `we starts` -> `we start`
- `that are less far apart` -> `that are much closer`
- `and finally comparing adjacent elements.` -> `and then finally compare adjacent elements, `

faheel: The following description would be better in my opinion:
> `An in-place comparison sort which starts by sorting pairs of elements far apart from each other, then progressively reducing the gap between elements to be compared. The time complexity depends on the gap sequence, which is floor(N/2^k) in this implementation.`

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    Time complexity
+    ----------------
+    O(N*log(N)^2), where N is the nubmer of elements.
```
faheel: For the gap sequence used, the time complexity is O(N^2), or to be precise, Theta(N^2).

faheel (changes requested):
Please make use of the utility functions in [SortingUtils.hpp](../blob/master/Sorting/SortingUtils.hpp). You can take help from other [sorting programs](../blob/master/Sorting) to see how and where to use these functions.

Hossam-Tarek:
please review this implementation now @faheel

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,62 @@
...
+            if (show_state)
+                display_state(values);
+        }
+    }
+}
+
+// Print the array elements.
+void print(const vector<int> &arr) {
+    for (int i = 0; i < arr.size(); ++i)
+        cout << arr[i] << " ";
```
faheel: This can now be removed.

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,62 @@
...
+void shell_sort (vector<int> &values, const bool &show_state) {
+    int n = values.size();
+
+    // We calculate the gap sequence in order of |n/2|, |n/4|,....
+    for (int gap = n/2; gap > 0; gap /= 2) {
+        for (int i = gap; i < n; ++i) {
+
+            int temp = values[i];
+            int j;
+            for (j = i; j >= gap && values[j - gap] > temp; j -= gap)
```
faheel: Use `and` and `or` instead of `&&` and `||` respectively.

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void shell_sort (vector<int> &values, const bool &show_state) {
+    int n = values.size();
+
+    // We calculate the gap sequence in order of |n/2|, |n/4|,....
+    for (int gap = n/2; gap > 0; gap /= 2) {
+        for (int i = gap; i < n; ++i) {
+
+            int temp = values[i];
+            int j;
```
faheel: Declare the variables `temp`, `i` and `j` outside the loops to prevent redeclaration on every iteration.

Sorting/ShellSort.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    vector<int> values(size);
+    get_input_values(values, size);
+    
+    bool show_state;
+    get_whether_to_show_state(show_state);
+
+    shell_sort(values, show_state);
+    display_state(values);
+}
```
faheel: `return 0;` is missing at the end.

faheel (changes requested):
A few more changes are required before this can be merged.

Hossam-Tarek:
what about now?
and do you have any reason of putting 'return 0' the compiler without it works fine.

faheel:
@Hossam-Tarek, looks good now! 👍 

And about `return 0;`, it is required, by standard, at the end of `main()` because the return code signals the OS that the program has exited successfully (`0` is success, non-0 is failure). This helps in debugging a software where lots of programs are involved, and their exit status may be checked to know whether or not a program executed successfully. If you don't write it, the program will return some random value, which is not helpful in any way.

Hossam-Tarek:
@faheel thank you for your help
this was my first contribution in GitHub

faheel:
Happy to help :smiley: . And congratulations on your first PR merge! :tada:

PR #53, opened 2017-05-11
Add Recursion Folder and Tower of Hanoi algorithm 

knotman90:
It also contains code for pretty printing the status of the tower.
refresh speed is controlled via sleeptime variable.

### DIFF
--- a/Recursion/TowerHanoi.cpp
+++ b/Recursion/TowerHanoi.cpp
@@ -1,9 +1,32 @@
+/*
+@Author Davide Spataro www.davidespataro.it
+
+Tower Of Hanoi
+-------------------
+Solves the tower of hanoi problem ( https://en.wikipedia.org/wiki/Tower_of_Hanoi )
+using a recursive approach.
+
+
+Time complexity
+---------------
+O(2^n),where n is the number of disks
+
+Space complexity
+----------------
+O(n), where n is the number of disks 
+(not considering the stack space for the required by recursion). 
+*/
+
+
+
 #include <iostream>
 #include<array>
 #include<vector>
 #include <stack>
 #include <unistd.h>
 using namespace std;
+
+
 
 int ndisks;
 int sleeptime = 9000;
@@ -22,7 +45,7 @@ void drawHanoi(auto& piles) {
   for (int  i = ndisks ; i >= 0; --i) {
     for (int j = 0; j < 3; j++) {
       printf(" | ");
-      if (piles[j].size() > i) {
+      if (piles[j].size() > (unsigned)i) {
         int n = piles[j][i] + 1;
         printNtimes('=', n);
         printNtimes(' ', sep[j] - n + 1);
@@ -40,7 +63,7 @@ void drawHanoi(auto& piles) {
 
 
 /**
-  Function that solves the towe of hanoi problem
+  Function that solves the tower of hanoi problem
 */
 void solveHanoi(int start, int end, int from, int to, auto& piles) {
   //which pile should we use as support pile?
@@ -66,7 +89,6 @@ void solveHanoi(int start, int end, int from, int to, auto& piles) {
 
 
 int main() {
-
 
   int np;
   cout << "How many disks?...\n";
--- a/Recursion/TowerHanoi.cpp
+++ b/Recursion/TowerHanoi.cpp
@@ -1,0 +1,88 @@
+#include <iostream>
+#include<array>
+#include<vector>
+#include <stack>
+#include <unistd.h>
+using namespace std;
+
+int ndisks;
+int sleeptime = 9000;
+int nmoves = 0;
+
+//prints the current status of the piles
+void drawHanoi(auto& piles) {
+
+  auto printNtimes = [](char c, int n) {
+    for (int i = 0 ; i < n; i++)
+      printf("%c", c);
+  };
+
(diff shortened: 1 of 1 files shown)


Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+
+int main() {
+
+  int np;
+  cout << "How many disks?...\n";
+  cin >> np;
+  np--;
+
+  vector<vector<int>> piles(3, vector<int>());
```
faheel: Since the number of piles are fixed (3), declare `piles` as:
```C++
array<vector<int>, 3> piles;
```

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include<array>
+#include<vector>
+#include <stack>
+#include <unistd.h>
+using namespace std;
+
+
+
+int ndisks;
+int sleeptime = 9000;
```
faheel: Please remove `ndisks` as a global variable. Also `sleeptime` can be removed as there is no need for the sleep functionality.

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+//prints the current status of the piles
+void drawHanoi(auto& piles) {
+
+  auto printNtimes = [](char c, int n) {
+    for (int i = 0 ; i < n; i++)
+      printf("%c", c);
+  };
+
+  std::array<int,3> sep= {ndisks, ndisks, ndisks};
```
faheel: No need of `std::`

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+
+
+int main() {
+
+  int np;
+  cout << "How many disks?...\n";
+  cin >> np;
+  np--;
```
faheel: You can change this to:
```C++
int num_discs;
cout << "Enter the number of discs : ";
cin >> num_discs;
```

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+      } else
+        printNtimes(' ', sep[j] + 1);
+
+      printf(" | ");
+    }
+    printf("\n");
+  }
+  printf("\n\n");
+
+  usleep(sleeptime);
```
faheel: `usleep()` can be removed as there is no need for the sleep functionality.

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+
+/**
+  Function that solves the tower of hanoi problem
+*/
+void solveHanoi(int start, int end, int from, int to, auto& piles) {
+  //which pile should we use as support pile?
+  int support = (from + to) % 4;
+  support = support == 0 ? 2 : support;
```
faheel: Can this be simplified? Also please add a comment to explain how/why a pile is selected as a support.

faheel (changes requested):
Sorry for the late review :sweat_smile:

Recursion/TowerHanoi.cpp
knotman90: If there is no need for usleep() then I think I can get rid of the drawHanoi function too. It would be unusable withouth a sleep cause it will just execute too fast.

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,138 @@
...
+using std::endl;
+using std::array;
+using std::cin;
+using std::vector;
+
+/*
+  Class HanoiTowerSolver. it exposes a simple interface for solving
+  hanoi tower problem.
+*/
+class HanoiTowerSolver {
```
faheel, line 35: I think that using a solver class is unnecessary. It was fine the way it was before:
- A function that recursively solves the Tower of Hanoi problem, and after each move calls:
- Another function that draws the current status of the 3 piles.

This way it's much simple to understand. The initialisation etc. can be done in `main()`.

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+
+Time complexity
+---------------
+O(2^n),where n is the number of disks
+
+Space complexity
+----------------
+O(n), where n is the number of disks
+(not considering the stack space for the required by recursion).
```
faheel, line 17: `(not considering the internal stack space used for recursion)`

Recursion/TowerHanoi.cpp
```diff
@@ -0,0 +1,138 @@
...
+using a recursive approach.
+
+
+Time complexity
+---------------
+O(2^n),where n is the number of disks
+
+Space complexity
+----------------
+O(n), where n is the number of disks
```
faheel, line 16: It would be better if you used uppercase variables while mentioning the time and space complexity as they are easier to spot while reading.
For eg., consider:
`... where n is the number ...`
vs
`... where N is the number ...`
Not much of a difference, but that's how it's written elsewhere in the repo.

faheel (changes requested):
Please use 4 spaces as one level of indentation.

faheel:

So, what do you say?

knotman90:
Sounds good to me =)

faheel:

PR #61, opened 2017-06-04
Adding Graph

deathcod:
Below are BFS and DFS, more to come

### DIFF
--- a/DataStructures/Graph/BreadthFirstSearch.cpp
+++ b/DataStructures/Graph/BreadthFirstSearch.cpp
@@ -1,0 +1,56 @@
+/* Breadth-first search/traversal
+ * iterates through all vertices in an undirected graph in breadth-first order
+ *
+ * Time-complexity: 
+ * O(V + E), where V is number of vertices and E is number of edges
+ *
+ * Space-complexity:
+ * O(V)
+ *
+ */
+
+#include <queue>
+#include "Graph.h"
+
+void breadthFirstSearch(Graph &graph, const int source_vertex) {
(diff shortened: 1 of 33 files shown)


DataStructures/Graph/BreadthFirstSearch.cpp
```diff
@@ -0,0 +1,56 @@
...
+                visited[itr.first] = true;
+                toVisit.push(itr.first);
+            }
+        }
+        toVisit.pop(); /* traverse to next vertex */
+    }
+    std::cout << "\n";
+}
+
+int main() {
```
ryoung9550, line 39: This test function should be renamed a different name then called by the main test.cpp

DataStructures/Graph/DepthFirstSearch.cpp
```diff
@@ -0,0 +1,86 @@
...
+            DepthFirstSearchTraversal(graph, itr.first);
+        }
+    }
+
+    //Colour the current vertex with BLACK to mark finish
+    colour[curr_vertex] = BLACK;
+    finishing_time[curr_vertex] = time = time +1; 
+}
+
+int main() {
```
ryoung9550, line 69: Same as the DFS cpp file, the function should be named something else so it can be tested. This would create multiple entry points when compiling.

ryoung9550 (changes requested):

faheel:
Sorry for not reviewing your PR back then. The project has been updated quite a lot, so if you're still interested in getting this PR merged, feel free to re-open it and add a comment stating that you are, and I'll help you in making the required changes. Sorry once again for replying so late, and thanks a lot for taking the time to contribute to the project.

Issue #59, opened 2017-05-29
Add documentation for all data structures and algorithmic functions


faheel:
Add a README to each directory containing proper documentation for algorithms or data structures in that directory.

sriram-25:
What all should the documentation contain? For eg. Time Complexity,Working of the Algorithm?
@faheel

faheel:
For now the documentation will not include the working details of the algorithm but only information about the "interface" of various functions, i.e. about:
- what the function does,
- its inputs,
- its output (return value), if any,

along with (wherever necessary),
- time complexity,
- space complexity.

The inputs, variables and output will have their type, name and a single line description. The time and space complexity will be of the best and worst cases (and average case too, if necessary).

For example, for [Kadane.cpp](https://github.com/faheel/Algos/blob/master/MaximumSubarray/Kadane.cpp) the documentation could look like:

---
## `maximumSubarray`
<description of what the function does in 1-2 lines>

### Input
- #### `values`
  Type: `const vector<int> &`

  <description in 1-2 lines>

### Output
- #### tuple of `maxSum`, `start`, `end`
  Type: `tuple<int, size_t, size_t>`

  <description of `maxSum` in 1-2 lines>
  <description of `start` in 1-2 lines>
  <description of `end` in 1-2 lines>

### Time complexity
- #### Worst case
  _O_(**N**), where **N** is the size of `values`
- #### Best case
  _O_(**N**), where **N** is the size of `values`

---

Markdown for the above documentation:
```
## `maximumSubarray`
<description of what the function does in 1-2 lines>

### Input
- #### `values`
  Type: `const vector<int> &`

  <description in 1-2 lines>

### Output
- #### tuple of `maxSum`, `start`, `end`
  Type: `tuple<int, size_t, size_t>`

  <description of `maxSum` in 1-2 lines>
  <description of `start` in 1-2 lines>
  <description of `end` in 1-2 lines>

### Time complexity
- #### Worst case
  _O_(**N**), where **N** is the size of `values`
- #### Best case
  _O_(**N**), where **N** is the size of `values`
```

yumyumbird:
I'd like to take care of that.

faheel:
Issue #74 is now closed

faheel:
This task can be automated using [Doxygen](http://www.stack.nl/~dimitri/doxygen/).

ghost:
I can definitely get start on this, if it's still open.

chhawip:
[...]

PR #62, opened 2017-06-04
Documentation

fakabbir:
Initial commit for Sorting Documentation.


faheel:
Documentation for a broad topic, such as sorting, searching, backtracking etc. is unnecessary as it would only have a general introduction, which can be found elsewhere, and contents - which are good to have but already exist in the [Wiki](https://github.com/faheel/Algos/wiki/Contents).

fakabbir:
The problem with maintaining such a repo is that we need to put a starting point. Since most of the algorithms are always available over wikis or competitive coding websites. 

The above was just a ticket to start a discussion since, I also have aimed to have such a repo in past. We need to comeup with a serious idea if our aim is to provide a common ground for such algorithms.

*What do you say ?*

faheel:
I do intend to add descriptions about the working of various algorithms, but for now my intention is to have documentation for various functions and user-defined data types, so that when they need to be re-used in other programs, it makes things easier for the user to know, for example, what a certain *class* is for, what are its data members for, what member functions are available and what input do they take and what output do they give etc.

I have given an example of what could be added for now in issue #59.

Issue #63, opened 2017-06-07
Bucket sort


ritwik12:
Hey, 
Bucket sort is not included . I would like to include it too.
Also according to issue #36 radix sort is still not there, I would like to add that too.

Rajiv2605:
Can I contribute to this? This would be first open source contribution and it would be great if you could tell me how to go about it.

ritwik12:
BucketSort is already contributed by me with PR #64 . You may work on Radix sort.

Rajiv2605:
Okay, I will do the Radix Sort.

Rajiv2605:
I have submitted a pull request named "Created Radix Sort" with the Radix Sort code. Please review it.

ritwik12:
@faheel  is the owner of this repository. He will review and merge it @Rajiv2605 .  
:)

ritwik12:
@faheel  is the owner of this repository, So only he can review and merge it @Rajiv2605 :)

ritwik12:
Closing being unappropriated.

PR #64, opened 2017-06-07
Create BucketSort.cpp

ritwik12:
Added BucketSort.cpp
Relates to issue #63.

### DIFF
--- a/Sorting/BucketSort.cpp
+++ b/Sorting/BucketSort.cpp
@@ -1,0 +1,72 @@
+/*
+    Bucket sort
+    ----------
+    Bucket sort, or bin sort, is a sorting algorithm that works by distributing the elements of an array into a number of buckets. 
+    Each bucket is then sorted individually, 
+    either using a different sorting algorithm, 
+    or by recursively applying the bucket sorting algorithm.
+
+    Time complexity
+    ---------------
+    O(N), where N is the number of elements.
+
+    Space complexity
+    ----------------
+    O(N), where N is the number of elements.
+    
+*/
+
+#include <iostream>
+#include <vector>
+
+#include "SortingUtils.hpp"
+
+using namespace std;
+
+void bucket_sort (vector<int>& values, const int order, const bool to_show_state)
+{
+    size_t i, j;
+    int m = 101;
+    int buckets[m]; //Create m empty buckets  
+    
+	
+
+    for (int i = 0; i < m; ++i)
+	buckets[i] = 0;
+	//Increment the number of times each element is present in the input array. 
+        //Insert them in the buckets
+	for (int i = 0; i < values.size(); ++i)
+	    ++buckets[values[i]];
+
+     	    //Sort using insertion sort and concatenate 
+		for (int i = 0, j = 0; j < m; ++j)
+		    for (int k = buckets[j]; k > 0; --k)
+		       values[i++] = j;
+		       
+                       if (to_show_state)
+                           display_state(values);
+
+
+}
+
+int main() {
+    size_t size;
+    get_input_size(size);
+
+    vector<int> values(size);
+    get_input_values(values, size);
+
+    int order;
+    string order_text;
+    get_order(order, order_text);
+
+    bool to_show_state;
+    get_whether_to_show_state(to_show_state);
+
+    bucket_sort(values, order, to_show_state);
+
+    cout << "\nThe values in " << order_text << " order are:\n";
+    display_state(values);
+
+    return 0;
+}
--- a/Sorting/BucketSort.cpp
+++ b/Sorting/BucketSort.cpp
@@ -28,15 +28,15 @@ void bucket_sort (vector<int>& values, const bool to_show_state)
 {
     array<int, MAX_BUCKETS > buckets ; //Create MAX_BUCKETS number of empty buckets  
     for (size_t i = 0; i < MAX_BUCKETS; ++i) 
-	buckets[i] = 0;
-	//Increment the number of times each element is present in the input array. 
-	//Insert them in the buckets
-	for (size_t i = 0; i < values.size(); ++i)  
-	    ++buckets[values[i]];
-	    //Sort using insertion sort and concatenate 
-	    for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
-		for (int k = buckets[j]; k > 0; --k) 
-		values[i++] = j;
+	 buckets[i] = 0;
+	 //Increment the number of times each element is present in the input array. 
+	 //Insert them in the buckets
+	 for (size_t i = 0; i < values.size(); ++i)  
+	     ++buckets[values[i]];
+	     //Sort using insertion sort and concatenate 
+	     for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
+		 for (int k = buckets[j]; k > 0; --k) 
+		 values[i++] = j;
 		       
                 if (to_show_state)
                    display_state(values);
--- a/Sorting/BucketSort.cpp
+++ b/Sorting/BucketSort.cpp
@@ -24,23 +24,25 @@
 using namespace std;
 const int MAX_BUCKETS = 99999; //  maximum number of Buckets
 
-void bucket_sort (vector<int>& values, const bool to_show_state)
-{
+void bucket_sort (vector<int>& values, const bool to_show_state) {
     array<int, MAX_BUCKETS > buckets ; //Create MAX_BUCKETS number of empty buckets  
+    
     for (size_t i = 0; i < MAX_BUCKETS; ++i) 
 	 buckets[i] = 0;
 	 //Increment the number of times each element is present in the input array. 
 	 //Insert them in the buckets
-	 for (size_t i = 0; i < values.size(); ++i)  
-	     ++buckets[values[i]];
-	     //Sort using insertion sort and concatenate 
-	     for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
-		 for (int k = buckets[j]; k > 0; --k) {
-		 values[i++] = j;
+	
+    for (size_t i = 0; i < values.size(); ++i)  
+	 ++buckets[values[i]];
+	
+    //Sort using insertion sort and concatenate 
+    for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
+	 for (int k = buckets[j]; k > 0; --k) {
+	 values[i++] = j;
 		       
-                if (to_show_state)
-                   display_state(values);
-		 }	 
+        if (to_show_state)
+            display_state(values);
+	 }	 
 }
 
 int main() {
--- a/Sorting/BucketSort.cpp
+++ b/Sorting/BucketSort.cpp
@@ -24,7 +24,7 @@
 using namespace std;
 const int MAX_BUCKETS = 99999; //  maximum number of Buckets
 
-void bucket_sort (vector<int>& values, const int order, const bool to_show_state)
+void bucket_sort (vector<int>& values, const bool to_show_state)
 {
     array<int, MAX_BUCKETS > buckets ; //Create MAX_BUCKETS number of empty buckets  
     for (size_t i = 0; i < MAX_BUCKETS; ++i) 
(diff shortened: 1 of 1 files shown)


ritwik12:
Anything about Pull Request??

ritwik12:
@faheel  I think you may have forgot to review.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,72 @@
...
+using namespace std;
+
+void bucket_sort (vector<int>& values, const int order, const bool to_show_state)
+{
+    size_t i, j;
+    int m = 101;
+    int buckets[m]; //Create m empty buckets  
+    
+	
+
```
faheel: Remove the extra empty lines here.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,72 @@
...
+    for (int i = 0; i < m; ++i)
+	buckets[i] = 0;
+	//Increment the number of times each element is present in the input array. 
+        //Insert them in the buckets
+	for (int i = 0; i < values.size(); ++i)
```
faheel: Either you're missing curly braces after the `for` loop or the indentation is improper.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <vector>
+
+#include "SortingUtils.hpp"
+
+using namespace std;
+
+void bucket_sort (vector<int>& values, const int order, const bool to_show_state)
+{
+    size_t i, j;
+    int m = 101;
```
faheel: Declare `m` as a global constant giving it some meaningful name along with a short comment describing it or its use.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,72 @@
...
+    int buckets[m]; //Create m empty buckets
```
faheel: Use `array<int, ...> buckets` instead.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,72 @@
+/*
+    Bucket sort
+    ----------
+    Bucket sort, or bin sort, is a sorting algorithm that works by distributing the elements of an array into a number of buckets.
```
faheel: Keep lines 80 characters in length at max.

faheel (changes requested):
Yes, I had 😅. But here goes:

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,74 @@
+/*
+    Bucket sort
+    ----------
+    Bucket sort, or bin sort, 
+    is a sorting algorithm that works by 
+    distributing the elements of an array into a number of buckets. 
+    Each bucket is then sorted individually, 
+    either using a different sorting algorithm, 
+    or by recursively applying the bucket sorting algorithm.
```
faheel: How about:
```
Bucket sort, or bin sort, is a sorting algorithm that works by distributing the
elements of an array into a number of buckets. Each bucket is then sorted
individually using insertion sort (or some other sorting algorithm, including
bucket sort itself).
```

ritwik12: That will also be good, I will do it

Sorting/BucketSort.cpp
ritwik12: @faheel You mean vector<int,...> buckets I guess

Sorting/BucketSort.cpp
ritwik12: @faheel  I can't find any problem in indentation here. And no curly braces are required as each for loop consists of single line only.
Help me If I am missing something.

Sorting/BucketSort.cpp
ritwik12: Here m is the maximum range of number .
when m = 101 , We cannot use any number more than 101.

I can do two things now -
     1. Make m = 999999999999 (some very large value) and declare it globally.
     2. I can declare int m   globally and get its value from the user by adding cin>>m at the first   line of int main()
Which one you prefer ? let me know

ritwik12: For now I am declaring m = 999999999 globally.

ritwik12:
@faheel All the changes you requested are done.
Please review again :)

Sorting/BucketSort.cpp
faheel: No, I meant [`array<int, ...>`](http://en.cppreference.com/w/cpp/container/array). Eg:
```C++
array<int, MAX_BUCKETS> buckets;
```

Sorting/BucketSort.cpp
faheel: I take it that `m` is the maximum number of buckets. So what you should do is globally declare it as a constant:
```C++
const int MAX_BUCKETS = 101;
```

Sorting/BucketSort.cpp
faheel: Lines 36-39 and 41-47 have extra indentation.

faheel:
@ritwik12 Click on `Files changed` tab at the top to see what changes remain to be made.
*Pro-tip:* Once you make all the requested changes, all comments will disappear.

Sorting/BucketSort.cpp
ritwik12: If MAX_BUCKETS = 101 ,then you will not be able to sort numbers greater than 101.
for example array = {5,2,100,500,400}
here the sorting will not give proper result because of the low bucket size.
So I will recommend some very large value for MAX_BUCKETS.
such ad MAX_BUCKETS = 9999999999

faheel:
Please resolve all warning before committing the code:
```
BucketSort.cpp: In function ‘void bucket_sort(std::vector<int>&, int, bool)’:
BucketSort.cpp:31:5: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
     for (int i = 0; i < MAX_BUCKETS; ++i)
     ^~~
BucketSort.cpp:35:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
  for (int i = 0; i < values.size(); ++i)
  ^~~
BucketSort.cpp:35:20: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  for (int i = 0; i < values.size(); ++i)
                  ~~^~~~~~~~~~~~~~~
BucketSort.cpp:35:2: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
  for (int i = 0; i < values.size(); ++i)
  ^~~
BucketSort.cpp:38:6: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
      for (int i = 0, j = 0; j < MAX_BUCKETS; ++j)
      ^~~
BucketSort.cpp:38:6: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
BucketSort.cpp:42:17: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
                 if (to_show_state)
                 ^~
BucketSort.cpp:29:12: warning: unused variable ‘i’ [-Wunused-variable]
     size_t i, j;
            ^
BucketSort.cpp:29:15: warning: unused variable ‘j’ [-Wunused-variable]
     size_t i, j;
               ^
BucketSort.cpp:27:50: warning: unused parameter ‘order’ [-Wunused-parameter]
 void bucket_sort (vector<int>& values, const int order, const bool to_show_state)
                                                  ^~~~~
```

ritwik12:
@faheel All changes requested are done finally.
I don't think the comments will disappear as I changed the whole file content at once by copying from my code at pc.
Please click on view file and see that everything is good to go now :)

faheel:
@ritwik12 Please resolve the above warnings.

ritwik12:
@faheel Done :+1: 
All warnings are removed :)

faheel:
@ritwik12 Lots of `misleading-indentation` still present in lines 31 to 42 :disappointed:
Also `order` is unused.

ritwik12:
@faheel I have solved the issue about order.
For "Lots of misleading-indentation still present in lines 31 to 42" - I have already corrected it.Please click on view file on the top right under file changed tab.
https://github.com/ritwik12/Algos/blob/71214354268ffcbc83c05928023da55cbb58a6e7/Sorting/BucketSort.cpp
I don't know why the indentation changes are not reflecting in the preview :(

faheel:
I think it's because you have used mixed tabs and spaces in your code. Convert all tabs to 4 spaces using your text editor's option and then fix the indentation.

ritwik12:
@faheel All good now :)
I think their is some problem with preview. It adjusts according to itself , not according to file.

Sorting/BucketSort.cpp
```diff
@@ -0,0 +1,65 @@
...
+	 //Insert them in the buckets
+	 for (size_t i = 0; i < values.size(); ++i)  
+	     ++buckets[values[i]];
+	     //Sort using insertion sort and concatenate 
+	     for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
+		 for (int k = buckets[j]; k > 0; --k) 
+		 values[i++] = j;
+		       
+                if (to_show_state)
+                   display_state(values);
```
faheel: This should be inside the inner loop.

faheel (changes requested):

faheel:
Lines 31-40 still have tabs :(
Which text editor do you use?

Sorting/BucketSort.cpp
ritwik12: Thanks for this :)

ritwik12:
@faheel I can't see any problem now in indentation. It is all good in the preview.
I used ubuntu text editor. But for making changes I am just editing the file on Github itself.

faheel:
Don't do that! Download either Atom, Sublime Text 3, or VS Code. They are all great text editors for coding with lots of useful features.

ritwik12:
@faheel Ohk sure, but for now please tell me where is the problem?

screenshot from 2017-07-08 18 22 23

faheel:
This is how it should look like:
```C++
void bucket_sort (vector<int>& values, const bool to_show_state) {
    array<int, MAX_BUCKETS> buckets{};  // create MAX_BUCKETS number of empty buckets  
    
    // increment the number of times each element is present in the input array. 
    // Insert them in the buckets
    for (size_t i = 0; i < values.size(); ++i)  
        ++buckets[values[i]];
    
    // sort using insertion sort and concatenate 
    for (size_t i = 0, j = 0; j < MAX_BUCKETS; ++j)  
        for (int k = buckets[j]; k > 0; --k) {
            values[i++] = j;
                    
            if (to_show_state)
                display_state(values);
        }	 
}
```

faheel:
On looking at what the code does more closely I realised that this isn't bucket sort but rather non-stable counting sort! :open_mouth: You and I both need to read this [Wikipedia article on bucket sort](https://en.wikipedia.org/wiki/Bucket_sort) more closely. Really, really sorry for not realising this earlier.

ritwik12:
@faheel yes I read that the first time I thought of creating the code.
My code does the same work as the definition says, It creates bucket and distribute the array values into those buckets and then sort each Bucket. That is why I am having MAX_BUCKETS .
Try to take MAX-BUCKETS = 100 and then try to sort the values {500,100,300,32446}
it will give you segmentation fault because it can only create 100 buckets at max that can not hold these values.
whereas counting sort is totally different thing .

ritwik12:
@faheel  Finally everything is according to the requirements now..

ritwik12:
https://en.wikipedia.org/wiki/Bucket_sort  Please see the Pseudo code and match it with my code. It is same.

faheel:
Umm, I don't think so.
In bucket sort you create buckets (whose number is mostly kept equal to the number of elements) which contain a range of numbers, not just the count of a single number (that's what counting sort, and your code does).
As you can see in the pseudo-code, there is a function called `msbits()` which determines the bucket into which an element will go in. After the elements are put into the right buckets, each bucket is internally sorted using any sorting algorithm, and then the contents of all the buckets are concatenated to get the sorted elements.
So the thing that differentiates bucket sort from counting sort is a function like `msbits()` that determines in which bucket should an element be put in. That's what's missing from your code.

ritwik12:
@faheel you are getting a lot confused.
see http://www.sanfoundry.com/cpluscplus-program-implement-bucket-sort/ I took help from here.
It will make it clear

faheel:
The code at the link you gave says that it's for bucket sort, but is not. It's simply counting sort as I mentioned previously.
Please read the answer to the Stack Overflow question [Radix sort vs Counting sort vs Bucket sort. What's the difference?](https://stackoverflow.com/q/14368392/5245129) to know the subtle differences between counting sort, radix sort and bucket sort. Hope this clears things for you :)

faheel:

PR #66, opened 2017-06-08
Add BinomialCoeffcients.cpp

eskeype:
This program computes (N choose K) using a recurrence relation and dynamic programming. I think that this program would be a good addition as it showcases basic dynamic programming, and because I feel as though dynamic programming implementations of this algorithm are hard to find.

### DIFF
--- a/NumberTheory/BinomialCoefficient.cpp
+++ b/NumberTheory/BinomialCoefficient.cpp
@@ -5,7 +5,7 @@
     and a dynamic programming approach to memoize (save) BinomialCoefficient values computed between calls
 
     Time complexity:
-    O(N*K) - cache[i][j] will be computed exactly once for each i<=N and j<=K
+    O(N*K) - cache[i][j] will be computed at most once for each i<=N and j<=K
 
     Space complexity:
     O(N*K) due to memoization (due to the 'cache' variable), and O(N) stack space (due to recursion)
--- a/NumberTheory/BinomialCoefficient.cpp
+++ b/NumberTheory/BinomialCoefficient.cpp
@@ -1,14 +1,17 @@
 /*
     Binomial Coefficient:
-    For a set containing N unique elements, find the number of subsets containing K elements (aka N choose K)
-    This approach uses the recurrence BinomialCoefficient(N,K) = BinomialCoefficient(N-1,K) + BinomialCoefficient(N-1,K-1),
-    and a dynamic programming approach to memoize (save) BinomialCoefficient values computed between calls
+    For a set containing N unique elements, find the number of subsets
+    containing K elements (aka N choose K)
+
+    This approach uses the following recurrence:
+    B(n, k) = B(n-1, k) + B(n-1, k-1)
+    and a dynamic programming approach to memoize values computed between calls
 
     Time complexity:
-    O(N*K) - cache[i][j] will be computed at most once for each i<=N and j<=K
+    O(N*K)
 
     Space complexity:
-    O(N*K) due to memoization (due to the 'cache' variable), and O(N) stack space (due to recursion)
+    O(N*K)
 */
 
 #include <iostream>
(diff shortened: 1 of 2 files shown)


NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    }
+
+    cout << "Enter the value for K: ";
+    cin >> k;
+
+    while (k < 0) {
+        cout << "Invalid value! K should be greater than 0.\n";
+        cout << "Enter the value for K: ";
+        cin >> k;
```
faheel, line 79: Also check that `K <= N`.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+    Binomial Coefficient:
+    For a set containing N unique elements, find the number of subsets containing K elements (aka N choose K)
+    This approach uses the recurrence BinomialCoefficient(N,K) = BinomialCoefficient(N-1,K) + BinomialCoefficient(N-1,K-1),
+    and a dynamic programming approach to memoize (save) BinomialCoefficient values computed between calls
+
+    Time complexity:
+    O(N*K) - cache[i][j] will be computed at most once for each i<=N and j<=K
+
+    Space complexity:
+    O(N*K) due to memoization (due to the 'cache' variable), and O(N) stack space (due to recursion)
```
faheel: Space complexity is just `O(N * K)`. Also, there's no need to mention the reason here.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
+/*
+    Binomial Coefficient:
+    For a set containing N unique elements, find the number of subsets containing K elements (aka N choose K)
+    This approach uses the recurrence BinomialCoefficient(N,K) = BinomialCoefficient(N-1,K) + BinomialCoefficient(N-1,K-1),
+    and a dynamic programming approach to memoize (save) BinomialCoefficient values computed between calls
+
+    Time complexity:
+    O(N*K) - cache[i][j] will be computed at most once for each i<=N and j<=K
```
faheel: There's no need to mention the reason here.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+    O(N*K) due to memoization (due to the 'cache' variable), and O(N) stack space (due to recursion)
+*/
+
+#include <iostream>
+#include <vector>
+using namespace std;
+
+const int FLAG = -1;
+
+int BinomialCoefficientHelper(int N, int K, vector<vector<int> > cache)//uses recurrence relation and fills cache to find binomial coefficient of N & K
```
faheel: Pass `cache` by reference to prevent creating copies of it during each recursive call.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+    //out now contains the binomial coefficient of N & K
+
+    cache[N][K] = out;//out is stored in the cache so it doesn't need to be recomputed if needed again
+    return out;
+}
+
+int BinomialCoefficient(int N, int K)
+{
+    vector< vector<int> > cache;//2d cache vector will be used to memoize BinomialCoefficient values between calls
```
faheel: Please remove the spaces between `<` and `vector` and the two `>`s. C++11 and later versions can differentiate between the operator `>>` and nested angle brackets for template parameters.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+    cache.resize(N + 1);
+
+    for(int i = 0; i < (N + 1); i++)//cache[i][j] is initialized to FLAG value (-1) at start
```
faheel: Change `i < (N + 1)` to `i <= N` for simplicity.

NumberTheory/BinomialCoefficient.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+const int FLAG = -1;
+
+int BinomialCoefficientHelper(int N, int K, vector<vector<int> > cache)//uses recurrence relation and fills cache to find binomial coefficient of N & K
+{
+    //recursive invariant: cache[N][K] either contains binomial coefficient of N & K or FLAG (-1) (it is impossible for a binomial coefficient to be negative)
+    if(cache[N][K]!= FLAG)
+        return cache[N][K];
+
+    int out;//will contain the value to be returned
```
faheel: Rename `out` to `result`, as that would be a better name for the variable.

faheel (changes requested):
Try to keep each line less than 80 chars long. 80 is a soft limit while the hard limit is 100 chars.

eskeype:
Thanks for checking out my code! I will make all the updates that you suggested, but I want to hear your thoughts on having BinomialCoefficients(N, K) return 0 when N < K. I understand that some textbooks take this as the convention

eskeype:
Ok, committed the changes you suggested. A few of my lines went over 80 characters (2 or 3 lines), and I added the functionality that BinomialCoefficients(N,K) = 0 when N<K instead of updating the testing function. If you disagree I can change the function appropriately.

faheel:
Yes, C(n, K) = 0, for n < k.

faheel:
@eskeype Can you also add the test for this? It will be similar to the one for Fibonacci, which can be found [here](https://github.com/faheel/Algos/blob/master/Test/NumberTheory/Fibonacci.test.cpp).

eskeype:
Ok I'll do that. Should I add the overflow test as well? If so, do you think I should have the function take and return signed integers? Or should I switch to unsigned or unsigned long?

faheel:
Yes, you should add overflow cases as well.
You can use `unsigned long long` (which is 64-bit unsigned integer) as that would increase the range to `18446744073709551615`, from the current range of `2147483647`.
Also, don't forget to change `FLAG` from `-1` to `0`.

eskeype:
Ok absolutely. I will have BinomialCoefficients take 'int' inputs and return ULL outputs.

eskeype:
Ok, I made the suggested changes, and I added in some test cases. I calculated the test values using a python library function for binomial coefficients.

For the overflow tests, I found inputs with output values greater than 2^64, and I tested to see if the the outputs were equal to python's outputs modulo 2^64 (which I hardcoded into the testing file)

eskeype:
By the way, I noticed that Fibonacci.test.cpp wouldn't compile on my machine. I believe that this can be fixed by adding '#define TEST' to Fibonacci.test.cpp.

faheel:
Thanks for pointing out the missing `#define TEST` for Fibonacci.test.cpp. To compile the test I used to pass `TEST` as a flag using `-D` option for g++, but I now realise that adding a define in the test file is a much better approach.

faheel:
@eskeype Thanks for your time and contribution on this :smiley:! And congratulations on your first PR merge :tada:

eskeype:
Thanks for all of the guidance!

PR #67, opened 2017-06-11
Add Extended Euclidean Algorithm

eskeype:
I saw that there was already a GCD algorithm, but I thought it could be nice to include an extended Euclidean algorithm as I know that some number theoretic & cryptographic algorithms make use of these coefficients. Tell me your thoughts!

### DIFF
--- a/NumberTheory/ExtendedEuclid.cpp
+++ b/NumberTheory/ExtendedEuclid.cpp
@@ -12,7 +12,7 @@
 
     Time Complexity
     ---------------
-    O(log10(K)), where K = min(A, B) is the smaller of the input integers A and B
+    O(log(K)), where K = min(A, B) is the smaller of the input integers A and B
 
     Space Complexity
     ----------------
@@ -20,7 +20,7 @@
 */
 
 #include<iostream>
-#include<vector>
+#include<array>
 
 using namespace std;
 
@@ -47,14 +47,14 @@ using namespace std;
 
 */
 
-vector<int> extended_euclidean(int a, int b)
+array<int,2> extended_euclidean(int a, int b)
 {
     //loop invariant: gcd(a,b) is the same at the end of each iteration
     //a = aM[0]*(original value of a) + aM[1]*(original value of b)
     //the above statement holds for b and bM as well
 
-    vector<int> aM = {1,0};
-    vector<int> bM = {0,1};
+    array<int,2> aM = {1,0};
+    array<int,2> bM = {0,1};
     while(b!=0)
     {
         aM[0] -= (a/b)*bM[0];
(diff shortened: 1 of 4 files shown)


faheel:
There are some changes I made to [BinomialCoefficients.cpp](https://github.com/faheel/Algos/blob/master/NumberTheory/BinomialCoefficient.cpp) related to code style (variable and function names, comment style etc.) which I thought of requesting in the reviews before, but then decided that it would be too time consuming to go over each of them.
Since now the changes have been made, it would be nice if you could take a look at them and implement those in this program too.

P.S. More on the code style can be found in the [C++ coding guidelines](https://github.com/faheel/Algos/blob/master/C%2B%2B_CODING_GUIDELINES.md) and in the [Google style guide](https://google.github.io/styleguide/cppguide.html).

eskeype:
Ok, I'll work on the edits. Thanks!

eskeype:
Ok, I updated the comments and I borrowed the user input function you added to BinomialCoefficient.cpp and added it to ExtendedEuclid.cpp. Let me know if I need anything else

NumberTheory/ExtendedEuclid.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    gcd(a, b) = M[0]*a + M[1]*b
+
+    This algorithm has a variety of uses - in particular, it can be used to
+    quickly compute multiplicative inverses in finite fields, as is often
+    needed for cryptographic algorithms
+
+    Time Complexity
+    ---------------
+    O(log10(K)), where K = min(A, B) is the smaller of the input integers A and B
```
faheel: Base of the logarithm doesn't matter while writing the complexity, and therefore doesn't need to be mentioned.
Just `O(log(K)), where ...` is sufficient.

NumberTheory/ExtendedEuclid.cpp
```diff
@@ -0,0 +1,113 @@
...
+    --------------
+    Uses the recurrence relation,
+        GCD(a, b) = GCD(b, a%b),
+    to find coefficients M[0] and M[1] such that 
+        GCD(a, b) = M[0]*a + M[1]*b
+
+    Return value
+    ------------
+    vector<int> M of length 2 which satisfies the equation above
+
```
faheel: Since there will always be exactly 2 values that the function would return, it would be better to return them in either a `tuple<int, int>` or an `array<int, 2>`.

faheel (changes requested):

eskeype:
I updated the complexity description to no longer include the log10, and I opted for array<int,2> instead of vector<int>. Thanks!

faheel:
@eskeype Thank you once again for your contribution :smiley:

PR #68, opened 2017-06-13
Added perfect number check algorithm

hefty-head:
Wrote a simple program that checks whether or not a number is perfect

### DIFF
--- a/NumberTheory/PerfectNumbers.cpp
+++ b/NumberTheory/PerfectNumbers.cpp
@@ -14,7 +14,7 @@
  
 typedef unsigned long long int ULL;
 
- ULL is_perfect(ULL num) {
+ bool is_perfect(ULL num) {
  	ULL sum = 1;
  	ULL divisor;
  	for(divisor = 2; divisor <= sqrt(num); divisor++)
--- a/NumberTheory/PerfectNumbers.cpp
+++ b/NumberTheory/PerfectNumbers.cpp
@@ -1,20 +1,20 @@
-/*
+ /*
     Perfect number:
-    prints up to N perfect numbers
+    prints up to N perfect numbers (a positive integer that 
+    is equal to the sum of its proper positive divisors)
     Time complexity:
-    O(N), where N is the number being checked 
+    O(sqrt(N)), where N is the number being checked 
     Space complexity:
     O(1), constant amount of space needed
 */
 
 #include <iostream>
-#include <math.h>
 
 using namespace std;
 
 bool is_perfect(int num) {
 	int sum = 1;
-	for(int divisor = 2; divisor <= sqrt(num); divisor++)
+	for(int divisor = 2; divisor <= num/2; divisor++)
 	{
 		if(num % divisor == 0)
 			sum += divisor;
@@ -35,4 +35,3 @@ int main() {
 			cout << num << endl;
 	}
 }
- 
--- a/NumberTheory/PerfectNumbers.cpp
+++ b/NumberTheory/PerfectNumbers.cpp
@@ -1,0 +1,35 @@
+/*
+    Perfect number:
+    Checks if a given number is perfect. Could be used to generate a 
+    sequence of perfect numbers
+    Time complexity:
+    O(N), where N is the number being checked 
+    Space complexity:
+    O(1), constant amount of space needed
+*/
+
+#include <iostream>
+
+using namespace std;
+
+ULL is_perfect(int num) {
+	int sum = 1;
+	for(int divisor = 2; divisor <= num/2; divisor++)
+	{
+		if(num % divisor == 0)
+			sum += divisor;
+	}
+	if(divisor == sum)
+		return 1;
+	return 0;
+}
+
+int main() {
+	int num;
+	cout << "Enter a number: ";
+	cin >> num;
+	if(is_perfect(num))
+		cout << num << " is perfect" << endl;
+	else 
+		cout << num << " is not perfect" << endl;
+}
--- a/NumberTheory/PerfectNumbers.cpp
+++ b/NumberTheory/PerfectNumbers.cpp
@@ -20,16 +20,16 @@ typedef unsigned long long int ULL;
  	for(divisor = 2; divisor <= sqrt(num); divisor++)
  	{
  		if(num % divisor == 0)
- 			sum += divisor;
+ 			sum += divisor + num / divisor;
  	}
- 	if(divisor == sum)
- 		return 1;
- 	return 0;
+ 	if(num == sum)
+ 		return true;
+ 	return false;
  }
  
  int main() {
  	ULL num;
- 	cout << "Enter a number: ";
+ 	cout << "Enter a positive integer less than 2^64 : ";
  	cin >> num;
  	if(is_perfect(num))
  		cout << num << " is perfect" << endl;
--- a/NumberTheory/PerfectNumbers.cpp
+++ b/NumberTheory/PerfectNumbers.cpp
@@ -1,7 +1,6 @@
 /*
     Perfect number:
-    Checks if a given number is perfect. Could be used to generate a 
-    sequence of perfect numbers
+    prints up to N perfect numbers
     Time complexity:
     O(N), where N is the number being checked 
     Space complexity:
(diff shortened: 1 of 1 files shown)


NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,35 @@
+/*
+    Perfect number:
+    Checks if a given number is perfect. Could be used to generate a 
+    sequence of perfect numbers
```
faheel: It would be good if you could add a 1-2 line description of what a perfect number is.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace std;
+
+ULL is_perfect(int num) {
+	int sum = 1;
+	for(int divisor = 2; divisor <= num/2; divisor++)
+	{
+		if(num % divisor == 0)
+			sum += divisor;
+	}
```
faheel: The divisors can be found in _O_(sqrt(N)) by checking only upto sqrt(N), and adding `divisor` and `num / divisor` to `sum`.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,35 @@
...
+    O(N), where N is the number being checked 
+    Space complexity:
+    O(1), constant amount of space needed
+*/
+
+#include <iostream>
+
+using namespace std;
+
+ULL is_perfect(int num) {
```
faheel: The return type should be `bool`.

faheel (changes requested):
It would be nice if can modify the program to print all perfect numbers upto a given number.

hefty-head (commented):
Will do, thanks

hefty-head:
Just updated the code, thanks for the notes

hefty-head (commented):
I accidentally committed the wrong code whoops

NumberTheory/PerfectNumbers.cpp
faheel: You have changed the complexity but have forgotten to update the code here.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,37 @@
...
+    O(1), constant amount of space needed
+*/
+
+#include <iostream>
+
+using namespace std;
+
+bool is_perfect(int num) {
+	int sum = 1;
+	for(int divisor = 2; divisor <= num/2; divisor++)
```
faheel: `num`, `sum` and `divisor` should be of type `unsigned long long` so that they can go upto 2<sup>64</sup>-1.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,37 @@
...
+	}
+	if(divisor == sum)
+		return 1;
+	return 0;
+}
+
+int main() {
+	int num;
+	cout << "The number of perfect numbers you want printed: ";
+	cin >> num;
```
faheel: Instead of asking for
> the number of perfect numbers to be printed

ask for
> the number upto which perfect numbers are to be found

as there are only 8 perfect numbers between 1 and 2<sup>64</sup>-1.

You would also need to change `num` to `limit` or something more meaningful, and also it's type to `unsigned long long` to support a higher range of values.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,37 @@
...
+	for(int count = 1; count < num; count++)
+	//start from one to prevent false positive with 0
```
faheel: This loop would also need to be modified accordingly.

faheel (changes requested):

eskeype:
Hi @faheel and @aaboaf, I was looking at this commit, and i noticed that the complexity of displaying all of the perfect numbers less than N is O(sqrt(N) * N). However, it we were to use a sieve method this could be done in O(N * log(N)) time. 

In particular, we could have a large int array (to eventually contain the sum of the divisors of each index) with index values up to N, and for each index i>=0 we could loop through all of the multiples of i in the array and add i to their value.

This sieve has space complexity of O(N), which might make it a less attractive option, as there are very few perfect numbers in the very large range of [1..(2^64) - 1]. However, this algorithm could be adapted to work as an efficient method to determine whether numbers are deficient or abundant as well.

faheel:
@eskeype Your method would definitely be more efficient in finding perfect numbers but at a very high cost of memory space. To find out the perfect numbers between 1 and 2<sup>27</sup> only, we would require about 1 GB of memory. That's a lot of memory for finding just 5 perfect numbers!

faheel:
@aaboaf I now realise that keeping the program as simple as it originally was would be the best, because finding all perfect numbers within a large range would either require a lot of memory if done very efficiently, or a lot a processing time if done not so efficiently.
So just determining whether a given number is perfect or not in _O_(√N) time and _O_(1) space is feasible even for very large values of N (upto 2<sup>64</sup>-1).

hefty-head:
Will do!

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+ 
+typedef unsigned long long int ULL;
+
+ ULL is_perfect(ULL num) {
+ 	ULL sum = 1;
+ 	ULL divisor;
+ 	for(divisor = 2; divisor <= sqrt(num); divisor++)
+ 	{
+ 		if(num % divisor == 0)
+ 			sum += divisor;
```
faheel: Should be
```C++
sum += divisor + num / divisor;
```

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+ 	}
+ 	if(divisor == sum)
+ 		return 1;
+ 	return 0;
```
faheel: Return `true` and `false` instead of `1` and `0`.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+ ULL is_perfect(ULL num) {
+ 	ULL sum = 1;
+ 	ULL divisor;
+ 	for(divisor = 2; divisor <= sqrt(num); divisor++)
+ 	{
+ 		if(num % divisor == 0)
+ 			sum += divisor;
+ 	}
+ 	if(divisor == sum)
```
faheel: The condition should be 
```C++
if (sum == num)
```

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+ 		return 1;
+ 	return 0;
+ }
+ 
+ int main() {
+ 	ULL num;
+ 	cout << "Enter a number: ";
```
faheel: Ask the user to `"Enter a positive integer less than 2^64 : "` to be precise.

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+ 	cin >> num;
+ 	if(is_perfect(num))
+ 		cout << num << " is perfect" << endl;
+ 	else 
+ 		cout << num << " is not perfect" << endl;
```
faheel: "is/is not **a** perfect **number**"

NumberTheory/PerfectNumbers.cpp
```diff
@@ -0,0 +1,38 @@
...
+     O(1), constant amount of space needed
+ */
+ 
+ #include <iostream>
+ #include <math.h>
+ using namespace std;
+ 
+typedef unsigned long long int ULL;
+
+ ULL is_perfect(ULL num) {
```
faheel: Return type should be `bool`.

faheel (changes requested):

NumberTheory/PerfectNumbers.cpp
faheel: @aaboaf You missed this :smile:

faheel (approved):

faheel:
Thank you for your time and contribution @aaboaf! And congratulations on your first PR merge :smiley:

PR #73, opened 2017-06-29
Created Radix Sort

Rajiv2605:
I have created the code for Radix Sort. I hope that you like it. Please give me suggestions if I could have done anything better as this is my first pull request and would love to learn more.


Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,103 @@
...
+*/
+
+#include <iostream>
+#include <vector>
+#include "SortingUtils.hpp"
+
+using namespace std;
+
+//No. of elements to be sorted
+size_t SIZE;
```
faheel: Rename `SIZE` to `size`.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+//order is set to 1 if the sorting is done in ascending order, -1 if the sorting is done in descending order
+int order;
+
+//These factors depend on whether the sorting is needed to be done in ascending or descending order
+int mult_factor = 1;
+int add_factor = 0;
```
faheel: Please don't make all variables global. Instead, declare them in the function where they're required and pass them to other functions when needed.
You should make `size`, `order`, `mult_factor` and `add_factor` local to `main()`.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,103 @@
...
+	int max_val = values[0];
+	for (int i=1; i < SIZE; i++) {
+		if (values[i] > max_val)
+			max_val = values[i];
+	}
+	return max_val;
+}
+
+//The function implements radix sort by repeatedly calling count sort for different scaling factors.
+void radix_sort(vector<int>& values, const size_t SIZE, const int order, const bool to_show_state) {
```
faheel: Also mention the time and space complexity of this function.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,103 @@
...
+//The setFactors function sets the above factors
+void setFactors() {
+    if(order==-1) {
+        mult_factor = -1;
+        add_factor = 9;
+    }
+}
+
+//This function performs count sort on digits extracted from the numbers based on the scaling factor
+void count_sort(vector<int>& values, const size_t SIZE, int I, const int order, const bool to_show_state) {
```
faheel: Also mention the time and space complexity of this function.

faheel (changes requested):
- Your code has mixed indentation. Please convert each indentation level to 4 spaces. You can configure your text editor to emit 4 spaces whenever you press the tab key.
- Mention the time and space complexity of those functions which are directly related to the algorithm being implemented, which are `count_sort` and `radix_sort` for this program. Click [here](https://github.com/faheel/Algos/blob/master/DataStructures/DisjointSet/DisjointSet.cpp#L28-L42) for an example.
- Don't make all variables global. Instead, declare them in the function where they're required and pass them to other functions when needed.

Rajiv2605:
@faheel I have done all the changes you suggested. Please review it.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+/*
+    count_sort
+    ----------
+    Performs Count Sort on the elements passed by the radix_sort function
+
+    Time complexity
+    ---------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
```
faheel: ~`This space is used for count sorting each key and temporarily creating a sorted array.`~
There's no need to mention this under time complexity.

faheel (commented):
Looks fine :+1: but I haven't tested it yet. Will do soon.

Rajiv2605:
@faheel I think now the code is fine. Please review it.

Rajiv2605:
@faheel Is the code fine?

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    //to_show_state will be set to true if the states are needed to be printed
+    bool to_show_state;
+    get_whether_to_show_state(to_show_state);
+
+    //These factors depend on whether the sorting is needed to be done in ascending or descending order
+    int mult_factor = 1;
+    int add_factor = 0;
+    setFactors(order, mult_factor, add_factor);
+
+    radix_sort(values, size, order, to_show_state, mult_factor, add_factor);
```
faheel: No need to pass `size` as `values.size()` will give the number of elements (in constant time).
Also, there's no need to pass `order` as `mult_factor` and `add_factor` are being passed.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    Time complexity
+    ---------------
+    O(D * N), where N is the number of keys, D is the number of digits in input integers.
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of elements and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void radix_sort(vector<int>& values, const size_t size, const int order, const bool to_show_state, const int mult_factor, const int add_factor) {
```
faheel: Change the signature to
```C++
void radix_sort(vector<int>& values, const int mult_factor, const int add_factor, const bool to_show_state)
```

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    for(int i = size - 1; i >= 0; i--) {
+        output[counter[add_factor + (mult_factor * ((values[i]/I) % 10))] - 1] = values[i];
+        counter[add_factor + (mult_factor * ((values[i]/I) % 10))]--;
+    }
+
+    for(int i = 0; i < size; i++)
+        values[i] = output[i];
+
+    if(to_show_state) {
+        cout << "Printing the current state..." << endl;
```
faheel: Simply:
```C++
if (to_show_state)
    display_state(values);
```
`display_state()` is defined in SortingUtils.hpp

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void count_sort(vector<int>& values, const size_t size, const int I, const int order, const bool to_show_state, const int mult_factor, const int add_factor) {
+    int counter[10] = {0};
+    int* output = (int*)malloc(size*sizeof(int));
```
faheel: Declaring `output` as
```C++
vector<int> output(values.size());
```
would be simpler and better.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of elements and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void radix_sort(vector<int>& values, const size_t size, const int order, const bool to_show_state, const int mult_factor, const int add_factor) {
+    int m = getMaxVal(values, size);
+    for (int i = 1; m/i > 0; i *= 10) {
+        count_sort(values, size, i, order, to_show_state, mult_factor, add_factor);
```
faheel: Use a better name for `i`.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    Time complexity
+    ---------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void count_sort(vector<int>& values, const size_t size, const int I, const int order, const bool to_show_state, const int mult_factor, const int add_factor) {
```
faheel: - Change the signature to:
  ```C++
  void count_sort(vector<int>& values, const int I, const int mult_factor, const int add_factor, const bool to_show_state)
  ```
- Use a better name for `I`.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    int counter[10] = {0};
+    int* output = (int*)malloc(size*sizeof(int));
+    for(int i = 0; i < size; i++)
+        counter[add_factor + (mult_factor * ((values[i]/I) % 10))]++;
```
faheel: Use a range-based for loop for simplification.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    for(int i = 1; i < 10; i++)
+        counter[i] += counter[i - 1];
+    for(int i = size - 1; i >= 0; i--) {
+        output[counter[add_factor + (mult_factor * ((values[i]/I) % 10))] - 1] = values[i];
+        counter[add_factor + (mult_factor * ((values[i]/I) % 10))]--;
+    }
+
+    for(int i = 0; i < size; i++)
+        values[i] = output[i];
```
faheel: No need of a loop here. Simply:
```C++
values = output;
```

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <vector>
+#include "SortingUtils.hpp"
+
+using namespace std;
+
+// The setFactors function sets the factors which depend on whether the sorting is needed to be done in ascending or descending order
+void setFactors(const int order, int& mult_factor, int& add_factor) {
+    if(order == -1) {
+        mult_factor = -1;
+        add_factor = 9;
```
faheel: This can be done in `main()` itself. No need for a separate function.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    ---------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void count_sort(vector<int>& values, const size_t size, const int I, const int order, const bool to_show_state, const int mult_factor, const int add_factor) {
+    int counter[10] = {0};
```
faheel: Declare `counter` as:
```C++
array<int, 10> counter{};
```
to be specific to C++.

faheel (changes requested):
Sorry for the late review :sweat_smile:. Here's a summary:
- No need to pass `size` as it can be obtained using `values.size()`
- Use C++ containers such as `std::array` and modern (C++1x) features like [range-based for loops](http://en.cppreference.com/w/cpp/language/range-for), rather than C-style arrays, `malloc()` etc.

Sorting/RadixSort.cpp
ritwik12: Don't forgot to use` #include<array>` while using `array<int, 10> counter{};`

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,120 @@
...
+    int* output = (int*)malloc(size*sizeof(int));
+    for(int i = 0; i < size; i++)
+        counter[add_factor + (mult_factor * ((values[i]/I) % 10))]++;
+    for(int i = 1; i < 10; i++)
+        counter[i] += counter[i - 1];
+    for(int i = size - 1; i >= 0; i--) {
+        output[counter[add_factor + (mult_factor * ((values[i]/I) % 10))] - 1] = values[i];
+        counter[add_factor + (mult_factor * ((values[i]/I) % 10))]--;
```
faheel: These expressions are hard to understand. Please simplify these.
Also `counter[add_factor + (mult_factor * ((values[i]/I) % 10))]` is used multiple times. I suggest using a function that returns the required reference, and adding comments explaining what the expression and the involved terms are.
Let me know if you need some help implementing the function.

Sorting/RadixSort.cpp
faheel: Yes :+1:

Rajiv2605:
@faheel I will do the suggested changes by tomorrow.

Sorting/RadixSort.cpp
Rajiv2605: Why do I need a function? I could rather use a local variable which would store the value of  `counter[add_factor + (mult_factor * ((values[i]/I) % 10))]` and then I would use that variable everywhere locally.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,113 @@
...
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void count_sort(vector<int>& values, const int extractor, const bool to_show_state, const int mult_factor, const int add_factor) {
+    array<int, 10> counter{0};
+    vector<int> output(values.size());
+    for(int i = 0; i < values.size(); i++)
+        counter[add_factor + (mult_factor * ((values[i]/extractor) % 10))]++;
```
faheel: Using a range-based for loop here will work as intended, so please add it back.
On line 38 however, since you are iterating the container in a reverse order, you cannot use a range-based for loop as it iterates from the beginning until the end.

Sorting/RadixSort.cpp
faheel: Since the value `add_factor + (mult_factor * ((values[i]/I) % 10))` depends on `values[i]`, or more specifically, just the index `i`, you cannot store it in a variable. You would either need to pass the value `values[i]`, or pass `values` and `i`, along with `add_factor`, `mult_factor` and `extractor` to a function which would return the required value.
As to why use a function for this at all, it is because it would reduce code repetition and make the code shorter and simpler to understand.
**NOTE:** Instead of creating a global function and passing it all the values (`values[i]`, `add_factor`, `mult_factor` and `extractor`), you can use a [lambda expression](http://en.cppreference.com/w/cpp/language/lambda) and pass only `values[i]` as the parameter to further simplify the expression.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    if(to_show_state) {
+        display_state(values);
+    }
+}
+
+//This function returns the maximum value in the vector
+int getMaxVal(const vector<int> values) {
+    int max_val = values[0];
+    for (int i = 1; i < values.size(); i++) {
```
faheel: Use [`max_element()`](http://en.cppreference.com/w/cpp/algorithm/max_element) from [`algorithm`](http://en.cppreference.com/w/cpp/algorithm) instead of this function.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,113 @@
...
+    O(D * N), where N is the number of keys, D is the number of digits in input integers.
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of elements and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void radix_sort(vector<int>& values, const bool to_show_state, const int mult_factor, const int add_factor) {
+    int m = getMaxVal(values);
+    for (int extractor = 1; m/extractor > 0; extractor *= 10) {
```
faheel: Rename `m` to `max_value`.

faheel (changes requested):

Rajiv2605:
@faheel I have done the changes.

Sorting/RadixSort.cpp
faheel: @Rajiv2605 You missed this

_Sent from my Samsung GT-I9500 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,109 @@
...
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void count_sort(vector<int>& values, const int extractor, const bool to_show_state, const int mult_factor, const int add_factor) {
+    array<int, 10> counter{0};
+    vector<int> output(values.size());
+    for(int value : values)
+    	counter[add_factor + (mult_factor * ((value/extractor) % 10))]++;
```
faheel: @Rajiv2605 I mentioned the following in a previous comment which you may have missed:

Since the value `add_factor + (mult_factor * ((values[i]/I) % 10))` depends on `values[i]`, or more specifically, just the index `i`, you cannot store it in a variable. You would either need to pass the value `values[i]`, or pass `values` and `i`, along with `add_factor`, `mult_factor` and `extractor` to a function which would return the required value.
As to why use a function for this at all, it is because it would reduce code repetition and make the code shorter and simpler to understand.
**NOTE:** Instead of creating a global function and passing it all the values (`values[i]`, `add_factor`, `mult_factor` and `extractor`), you can use a [lambda expression](http://en.cppreference.com/w/cpp/language/lambda) and pass only `values[i]` as the parameter to further simplify the expression.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,109 @@
...
+    O(D * N), where N is the number of keys, D is the number of digits in input integers.
+
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of elements and B is the base for representing the numbers(10 in case of decimal system)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+void radix_sort(vector<int>& values, const bool to_show_state, const int mult_factor, const int add_factor) {
+    int max_value = getMaxVal(values);
+    for (int extractor = 1; max_value/extractor > 0; extractor *= 10) {
```
faheel: Can you add a comment here mentioning what `extractor` is?

faheel (changes requested):

Sorting/RadixSort.cpp
Rajiv2605: @faheel Won't adding a lambda expression would make the code look more complex? Lambda would basically return the expression `add_factor + (mult_factor * ((values[i]/I) % 10))` in-line. Rather, we could just let the expression be there like it is right now. By variable, I meant, inside the loop, we could have something like this:
```
for(int value : values) {
    int x = add_factor + (mult_factor * ((values[i]/extractor) % 10))
    counter[x]++;
}
```

faheel: The above would be fine only if the expression was being used once, but it is used 3 times in this function, and so to reduce repetition (following the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)) we should use a function that returns the desired value. But to a function you would need to pass lots of variables to get the desired value. That is why I suggest using a lambda expression as the following:
```C++
int my_lambda = [&] (const int& value) {
    return add_factor + mult_factor * ((value / extractor) % 10);
};
```
Usage example:
```C++
for (const int& value : values)
    counter[my_lambda(value)]++;

...

for (int i = values.size() - 1; i >= 0; i--) {
    output[counter[my_lambda(values[i])] - 1] = values[i];
    ...
}
```
See, it's not at all complex! :smiley:

P. S. Remember to give the lambda a more meaningful name.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <iostream>
+#include <vector>
+#include <array>
+#include <algorithm>
+#include "SortingUtils.hpp"
+
+using namespace std;
+
+// The setFactors function sets the factors which depend on whether the sorting is needed to be done in ascending or descending order
+
```
faheel: This comment should be removed.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,112 @@
...
+/*
+    count_sort
+    ----------
+    Performs Count Sort on the elements passed by the radix_sort function
+
+    Time complexity
+    ---------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
```
faheel: Since the value of `B` is a constant (=10), the time and space complexity of `count_sort()` and the space complexity of `radix_sort()` would be,
**_O_**(N + 10)
= **_O_**(N), where N is the number of keys (values)

faheel (changes requested):

Rajiv2605:
@faheel I have done all the changes:+1: I guess. My compiler was throwing a lot of errors when I had declared the variable equated to the lambda expression as `int`. But, when I changed it to `auto` type, it worked! What would have caused the problem?

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,117 @@
...
+    Space complexity
+    ----------------
+    O(N + B), where N is the number of keys and B is the base for representing the numbers(10 in case of decimal system)
+    Hence, the above expression becomes O(N + 10) = O(N)
+    This space is used for count sorting each key and temporarily creating a sorted array.
+*/
+
+auto counter_accessor = [] (const int value, const int mult_factor, const int add_factor, const int extractor) {
+    return add_factor + (mult_factor * ((value/extractor) % 10));
+};
```
faheel: Have this local to `count_sort()` so that you wouldn't need to pass `mult_factor`, `add_factor` and `extractor` as arguments, they will automatically be captured as references if you declare the lambda as in the example I gave.

Sorting/RadixSort.cpp
faheel: @Rajiv2605 You missed this.

faheel (changes requested):

Sorting/RadixSort.cpp
faheel: @Rajiv2605 You missed this.

faheel:
About the lambda's type, I don't know why you were getting errors on `int` as there's nothing wrong with it. Did you try with `-std=c++14` flag?

faheel:

So, what do you say?

Sorting/RadixSort.cpp
Rajiv2605: I have done it in the next line.

Rajiv2605:
:+1: I'll wait for you to finish your PR. Let me know when you finish it.

Sorting/RadixSort.cpp
faheel: All the explanation about the complexity is not needed here. Simply write
```C++
Time complexity
---------------
O(N), where N is the number of keys.

Space complexity
----------------
O(N), where N is the number of keys.
```
Also elsewhere.

Sorting/RadixSort.cpp
faheel: You probably missed this.
Click on the `Files changed` tab for previous details.

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,117 @@
...
+    ----------
+    An efficient, key-index-based, comparision based sorting algorithm.
+    It sorts based on the keys obtained on each index of the numbers to be sorteed.
+*/
+
+#include <iostream>
+#include <vector>
+#include <array>
+#include <algorithm>
+#include "SortingUtils.hpp"
```
faheel: This should now be `#include "sorting/utils.hpp"`

Sorting/RadixSort.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    values = output;
+
+    if(to_show_state) {
+        display_state(values);
+    }
+}
+
+//This function returns the maximum value in the vector
+int getMaxVal(const vector<int> values) {
```
faheel: Take `values` by reference to avoid unnecessary copies of it being made.

faheel (changes requested):
- The file should now be placed in `C++/source/sorting/` with the name `radix_sort.cpp`.
- Take all `const` arguments as references to avoid unnecessary copies of them being made.

C++/source/sorting/radix_sort.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    Time complexity
+    ---------------
+    O(N), where N is the number of keys.
+
+    Space complexity
+    ----------------
+    O(N), where N is the number of keys.
+*/
+void count_sort(vector<int>& values, const int extractor, const bool to_show_state, const int mult_factor, const int add_factor) {
```
faheel: All of these parameters should taken as references too.

C++/source/sorting/radix_sort.cpp
```diff
@@ -0,0 +1,113 @@
...
+    array<int, 10> counter{0};
+    vector<int> output(values.size());
+
+    auto counter_accessor = [] (const int value) {
```
faheel: Capture list should be `[&]`

faheel (changes requested):

Rajiv2605:
@faheel Is it fine now?

faheel:
Oh no, it doesn't seem to work :disappointed:. With the input `3 6 2 0 -5 4 -8 -6 4 1 7 5 8 6 7 5` (size 16) it gives a seg fault when sorting in increasing order and gives the wrong values when sorting in decreasing order.

Rajiv2605:
@faheel Oh... :confused: It worked with many inputs I tried... I'll check... :disappointed:

Rajiv2605:
@faheel The code is unable to locate `sorting/utils.cpp`. Also, the reason you got error is that Radix Sort's general algorithm, i.e., the one I have implemented doesn't sort negative numbers. The code I had pushed which used `SortingUtils.hpp` works perfectly for non-negative numbers.

faheel:
Ah, right, it will only sort positive values. My bad! Please mention this fact in the description comment.

Regarding "not being able to locate `sorting/utils.hpp`", use the command **`make all`** from the `C++` folder to compile the sources. It's much more convenient and gives no issues. More details can be found in the [Compiling](../tree/master/C%2B%2B#compiling) section of the README.

Rajiv2605:
@faheel I have added it to the comments:v:

faheel:
@Rajiv2605 Thanks a lot for your time and contribution! :smiley:

Rajiv2605:
Thanks!

Issue #75, opened 2017-07-05
Fast Power Algorithm 


ahmad307:
Fast Power algorithm isn't present in "Number Theory" folder.  It's an algorithm for calculating exponentiation in O(log(N)) time.
I'd like to contribute by adding it.

faheel:
It would be nice having it. Go ahead!

Issue #41, opened 2017-04-04
Unit tests


faheel:
Add unit tests for the following programs:
- [x] N-Queens
- [x] Kadane
- [x] LCS
- [x] KMP
- Searching
	- [x] Common test for linear search and binary search 
	- [ ] Ternary search* 
- Sorting
	- [x] Common test
	- [ ] Radix sort

\* Ternary search currently has a unit test but it needs to be refactored.

faheel:
Read about [Adding unit tests for C++ code](../blob/master/C++/UNIT_TESTS.md) for step-wise directions.

faheel:
Issue #74 is now closed

alxmjo:
All of the **searching** algorithms currently have unit tests, but they don't have a _common_ unit test. However, there is a common test for **sorting**, which may be useful for implementing a common searching unit test.

alxmjo:
@faheel Regarding a common test for searching algorithms, is it your belief that this should include ternary search? The interface and use of ternary search differs from the other two existing search algorithms (linear and binary), which makes me wonder if it would be more straightforward to have a common test for binary and linear (and similar algorithms which may be added later) and then a separate test for ternary search. Not as clean as the sorting implementation I added earlier, but maybe a better fit for the objective at hand.

faheel:
@alxmjo Having a common unit test for linear and binary search and a separate one for ternary search would be best.

ghost:
Hey, I'd love to give this a crack. Let me know if the issue is still open for any left over algorithms.

alxmjo:
@safderareepattamannil If you're still interested, we're still missing a common test for searching algorithms. Have a look at the common test for sorting algorithms for an idea of how you might implement it.

beardbytes:
@alxmjo Hey, I am interested to give it a try !

amrii1:
Hello,
Is still any unit tests that needs to be done

alxmjo:
@amrii1 Yes! See the post at the top of the thread. Currently still need a common test for linear and binary search, and then standalone test files for ternary search and radix sort. Please create a new issue if you have questions or would like to discuss implementation. 🙂

abdulwahab-alobaid-2191115290:
hello the unit testing for the searching algorithms is checked in the readme.md , so is it still open or not ?
i am planning to work on that

faheel:
[...]

PR #76, opened 2017-07-05
Added Fast Power in Number Theory

ahmad307:
This is my first PR, please tell me if there are any issues to be fixed.
Relates to issue #75.


NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+Space Complexity
+------------------
+O(log(N)) where N is the power the number is raised to.
+
+*/
+#include <iostream>
+
+using namespace std;
+typedef long long ll;
```
faheel: Use `unsigned long long` for double the range.

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+//Function that returns x raised to the power of y
+ll fastPower (ll x,ll y)
```
faheel: Please use more meaningful variable names like `base` and `exponent` instead of just `x` and `y`.
Also, rename the function to `fast_power` to adhere to the guidelines.

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+<Fast Power>
```
faheel: Please remove the angle brackets (`<` and `>`) and indent the block by one level (4 spaces).

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+//Testing the function
+int main()
+{
+    int base,power;
+    cout<<"Enter the number and the power it's raised to:"<<endl;
+
+    cin>>base>>power;
+
+    cout<<fastPower(base,power);
```
faheel: Add a trailing `"\n"` so that the prompt on terminals is displayed in the next line.

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
...
+Fast Power is an optimized algorithm to compute exponentiation in a short time.
+
+
+Time Complexity
+-----------------
+O(log(N)) where N is the power the number is raised to.
+
+Space Complexity
+------------------
+O(log(N)) where N is the power the number is raised to.
```
faheel: ~`where N is the power the number is raised to`~
More simply, `where N is the exponent.`

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+<Fast Power>
+--------------
+Fast Power is an optimized algorithm to compute exponentiation in a short time.
+
```
faheel: It would be helpful for others if you could also mention in 1 or 2 lines more what the algorithm does.

faheel (changes requested):
It would be great if you also add tests for this, similar to the ones found [here](../blob/master/Test/NumberTheory). Let me know if you need more information regarding it.

ahmad307:
Regarding adding tests. Shall I add some sample tests commented?
Or shall I make them this way:

```
TEST_CASE("Base cases", "[fast_power]") {
   REQUIRE(fast_power(2,2) == 4);
   REQUIRE(fast_power(2,4) == 16);
}
```

faheel:
Like the example you gave.

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,67 @@
...
+    cout<<"Enter the number and its exponent:"<<endl;
+
+    cin>>base>>exponent;
+
+    cout<<endl<<fast_power(base,exponent)<<endl;
+    
+    //Sample Tests
+    TEST_CASE("Base cases", "[fast_power]") {
+        REQUIRE(fast_power(2,2) == 4);
+        REQUIRE(fast_power(2,4) == 16);
```
faheel: I think you didn't exactly understand how to add tests, but no worries :smile:! Read the second comment in issue #41 to know the exact steps in detail.

ahmad307:
I have now added a test file for fastPower.cpp, and edited fastPower.cpp to be compatible.

Test/NumberTheory/fastPower.test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+TEST_CASE("Normal cases", "[fast_power]") {
+    REQUIRE(fast_power(3,4) == 81);
+    REQUIRE(fast_power(7,9) == 40353607);
+    REQUIRE(fast_power(15,10) == 576650390625);
+}
+
+TEST_CASE("Overflow cases", "[fast_power]") {
+    REQUIRE(fast_power(2,100) == 0);
+    REQUIRE(fast_power(10,99) == 0);
```
faheel: `fast_power(2, 100)` should be `68719476736` (see [this](http://www.wolframalpha.com/input/?i=2%5E100+mod+(2%5E64-1))) and `fast_power(10, 99)` should be `4440381706574496940U` (see [this](http://www.wolframalpha.com/input/?i=10%5E99+mod+(2%5E64-1))). Suffix `U` is for `unsigned`.

NumberTheory/FastPower.cpp
```diff
@@ -0,0 +1,54 @@
...
+        return (base*base);
+    }
+}
+
+#ifndef FAST_POWER_TEST
+
+int main()
+{
+    //Testing the function
+    int base,exponent;
```
faheel: ~`int`~ `ull`

Test/NumberTheory/fastPower.test.cpp
```diff
@@ -0,0 +1,24 @@
+#define FAST_POWER_TEST
+#define CATCH_CONFIG_MAIN
+
+#include "../catch.hpp"
+#include "../../NumberTheory/FastPower.cpp"
+
+
+TEST_CASE("Base cases", "[fast_power]") {
+    REQUIRE(fast_power(2,2) == 4);
+    REQUIRE(fast_power(2,4) == 16);
```
faheel: Under "Base cases", check the values of the following:
* `fast_power(0, 0)` is `undefined` or [`NaN`](http://en.cppreference.com/w/cpp/numeric/math/nan). Comment this case here and handle it properly in the code.
* `fast_power(0, 1)` should be `0`
* `fast_power(1, 0)` should be `1`
* `fast_power(1, 1)`  should be `1`

faheel (changes requested):
Well done on adding the tests! :+1:
However there are a few issues I have mentioned in the review. Also, the `f` in the name of the test file should be capital.

Test/NumberTheory/fastPower.test.cpp
ahmad307: I think you meant ``` fast_power(2 ,100) ``` in the first one.
Anyway, regarding large numbers, to fix the output being ```0``` we can output them modulo a large number such as ```1000000000007``` , but not a huge number like ```2^64```. Where ```fast_power(2, 100)``` will be ```40955801449``` .
Regarding the last statement, do you mean changing ```ull``` to ```Ull``` ?

faheel: Yes, I meant `fast_power(2, 100)` in the first one. Fixed it.

I don't think you need to output modulo a specific number. The result will automatically be modulo `MAX_ULL`, which is 2<sup>64</sup>-1.

No, `ull` is fine. I meant you need to add the suffix `U` to unsigned numbers greater than 2<sup>63</sup>-1 to prevent warnings.

ahmad307: Aha! Sorry for the misunderstanding.

Test/NumberTheory/FastPower.test.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+
+TEST_CASE("Base cases", "[fast_power]") {
+    //REQUIRE(fast_power(0,0) == undefined);
+    REQUIRE(fast_power(0,1) == 0);
+    REQUIRE(fast_power(1,0) == 1);
+    REQUIRE(fast_power(1,1) == 1);
+    REQUIRE(fast_power(2,2) == 4);
+    REQUIRE(fast_power(2,2) == 4);
+    REQUIRE(fast_power(2,4) == 16);
```
faheel: The last 2 can be added under "Normal cases". Also the (2, 2) case has a duplicate.

ahmad307:
Requested changes are all done. Are there any other issues to be fixed?

Test/NumberTheory/FastPower.test.cpp
```diff
@@ -0,0 +1,28 @@
...
+    REQUIRE(fast_power(3,4) == 81);
+    REQUIRE(fast_power(7,9) == 40353607);
+    REQUIRE(fast_power(15,10) == 576650390625);
+}
+
+TEST_CASE("Overflow cases", "[fast_power]") {
+    REQUIRE(fast_power(2,100) == 68719476736);
+    REQUIRE(fast_power(10,99) == 4440381706574496940U);
```
faheel: Both these tests are failing, and I think the reason is that at some step of the calculation, the multiplication produces the value 2<sup>64</sup>, which for a 64-bit unsigned integer (`unsigned long long`) is 0, and so the final value also becomes 0 (as anything multiplied by 0 is 0).
This _could_ be solved by taking the modulus of the values at each step, like you suggested previously, but the value of that (prime) number should be as close as possible to 2<sup>64</sup>-1, to get exact results for a larger range of values.

faheel (changes requested):

Test/NumberTheory/FastPower.test.cpp
ahmad307: I have tried returning the result modulo several prime numbers. The number ```1000000000007``` seems to be suitable, as a number bigger than it will result in ```fast_power(2,100) ``` being greater than ```2^100 mod (2^64-1)```.
```fast_power(2,100)``` now returns ```40955801449```.
However, cases like ```fast_power(33,33)``` still cause overflow.
What do you think?

Comment (faheel) [reply Test/NumberTheory/FastPower.test.cpp]: I thought about this for a while and I think what we need to do is check how many digits the result is going to take. That would be done as follows:
```
digits_required = floor(exponent * log10(base)) + 1
```
If `digits_required > 19`, then use modulo `1000000000007` and tell the user about this too. Otherwise, we can directly use the result.

**Comment by faheel** [comment]

So, what do you say?

File: Test/NumberTheory/FastPower.test.cpp (hunk 12 above)
Comment (ahmad307) [reply Test/NumberTheory/FastPower.test.cpp]: Great idea!
but I want to note that we apply modulus ```1000000000007``` every time we return a value in ```fast_power()``` to prevent overflow in the function.
So we may create two functions, one that uses ```%1000000000007``` and one that doesn't. But to prevent this, I added a third parameter to the function called "mod" where we apply ```%1000000000007``` if ```digits_required > 19``` or else apply ```%10^18``` which is a number that won't affect the output result.
This is how the code is right now:
```
if (digits_required > 19)
        {
            cout<<endl<<fast_power(base,exponent,1000000000007)<<endl;
            cout<<"*The output is modulo 10^12+7"<<endl;
        }
        else
        {
            cout<<endl<<fast_power(base,exponent,1000000000000000000)<<endl;
        }
```
Shall I commit with the final changes so we can close the PR and put "FastPower" behind our backs? if that's not the most convenient thing to do right now I have no problem.

Comment (faheel) [reply Test/NumberTheory/FastPower.test.cpp]: Yes, passing a third parameter is the right way instead of having 2 similar functions. But use `ULLONG_MAX` as the mod value in the `else` case instead of 10<sup>18</sup>.

Comment (faheel) [reply Test/NumberTheory/FastPower.test.cpp]: **10<sup>12</sup>+7 is not prime!** So the modulo may be 0 in some cases. **Use 10<sup>9</sup>+7 instead.**

File: NumberTheory/FastPower.cpp [hunk 13]
```diff
@@ -16,23 +16,24 @@
...
 */
 #include <iostream>
+#include <cmath>
 
 using namespace std;
 typedef unsigned long long ull;
 
 //Function that returns base raised to the exponent
-ull fast_power (ull base,ull exponent)
+ull fast_power (ull base,ull exponent,ull mod)
```
faheel: Give `mod` the default value `ULLONG_MAX` so that it doesn't need to be passed to the function if modulus of the result is not required.

faheel (changes requested):
One last change and then we're good to merge!

Test/NumberTheory/FastPower.test.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include "../catch.hpp"
+#include "../../NumberTheory/FastPower.cpp"
+
+
+TEST_CASE("Base cases", "[fast_power]") {
+    //REQUIRE(fast_power(0,0) == undefined);
+    REQUIRE(fast_power(0,1,ULLONG_MAX) == 0);
+    REQUIRE(fast_power(1,0,ULLONG_MAX) == 1);
+    REQUIRE(fast_power(1,1,ULLONG_MAX) == 1);
```
faheel: `ULLONG_MAX` should be removed from the tests too. You missed this.

ahmad307: Sure! Sorry for this.

faheel (approved):

faheel:
Congratulations on your first PR merge! :tada: And thanks for contributing to the project :smiley:

ahmad307:
Thank you! 😃  Was great working with you 😄

Issue #74, opened 2017-07-02
Continuous integration


faheel:
## To-do
- [x] Reorganise the project
- [x] Write a Makefile to automate compilation, building and testing
- [x] Integrate with Travis CI and AppVeyor

### References
- https://github.com/faheel/Sample-Cpp-CI

PR #78, opened 2017-07-06
Add dijkstra algorithm

kapildeshpande:


faheel:

So, what do you say?

kapildeshpande:
Ok no problem.

C++/source/Graph algorithms/Dijkstra_algorithm.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include<vector>
+#include<set>
+#include<climits>
+using namespace std;
+
+#define MAX 10001
+
+vector < pair <int,int> > adj[MAX];
+bool visited[MAX];
+int dist[MAX];
```
faheel: Please make these containers local to `main()` and pass them by reference to whichever functions need them.

Also, you can declare `visited` and `dist` as `vector<bool>` and `vector<int>` respectively, which would remove the need for using the constant `MAX`.

C++/source/Graph algorithms/Dijkstra_algorithm.cpp
```diff
@@ -0,0 +1,84 @@
...
+				dist[adj[x][i].second] = dist[x] + y;
+				s.insert(make_pair(dist[adj[x][i].second],adj[x][i].second));
+			}
+		}
+
+	}
+}
+
+int main () {
+	int no_of_vertex,no_of_edges,source;
```
faheel: Use `num_` instead of `no_of_`, since it is a more widely used convention.

faheel (changes requested):
Thanks for adding Dijkstra's algorithm; it has been long overdue :sweat_smile:. It would be great if you could also add unit tests for it with a few test cases. The details for it can be found in [UNIT_TESTS.md](https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md).

Also, please rename the file to just `dijkstra.cpp` and its parent directory to `graph`.

kapildeshpande:
Ok no problem

faheel:

Issue #91, opened 2017-07-16
Priority Queue implementation using heaps.


kuldipkumark:
I would like to contribute the priority queue implementation using heaps.
With operations like insert, getmaxpriority, delmaxpriority.

faheel:
Sure, that would be good! But please wait until issue #74 is closed.

faheel:
Issue #74 is now closed

rsarky:
Can I take up this issue if noone is working on it?

PR #92, opened 2017-07-24
Priority Queue 

kuldipkumark:
Priority Queue using heap data structure added in C++/test/data_structures directory under #91 issue.


faheel (changes requested):
@Kuldip397 Thank you for contributing! There are a few things I would like to point out:
- A data structure should be implemented as a `class` so that it can be used in a way similar to STL containers. Eg. usage in an external program:
```C++
#include "data_structures/priority_queue.hpp"

...

PriorityQueue pq;
pq.insert(10);
pq.insert(5);
pq.insert(15);

while (!pq.is_empty()) {
    cout << pq.max();
    pq.remove();
}

...
```
- Mention it in the program whether the priority will be highest for the largest value or the smallest value.

- The code should be placed under the `include` directory, not `test`. The class should be declared in `priority_queue.hpp` and its implementation should be in `priority_queue.cpp` (both under `include/data_structures/`). The unit tests should be defined in `test/data_structures/priority_queue.test.cpp`.

kuldipkumark:
Thanks for your review. I'll make the changes as according to your suggestions.
Can you please make me clear, what I'm supposed to do with unit tests which is going to be defined in test/data_structures/priority_queue.test.cpp.

faheel:
In the unit test, you can insert and remove elements from the priority queue and after a few insertions and removals check that the `max()` returns the value that's expected. You can have 3-4 test cases with different values being inserted and removed, in each of which you should check that the `max()` returns the correct value 5 or more times. Also, check for edge cases like inserting the same value, or removing a value that's not present etc.

To know what changes you would need to do to add the unit test, read [Adding unit tests for C++ code](https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md).

C++/test/data_structures/PriorityQueue.test.cpp
```diff
@@ -0,0 +1,78 @@
+#define PRIORITY_QUEUE_TEST
+#define CATCH_CONFIG_MAIN
+
+#include "../../include/third_party/catch.hpp"
+#include "../../include/data_structures/PriorityQueue.cpp"
```
faheel: You can remove the prefix `../../include/` from the include path as the `include` directly is already added to the include path in the Makefile.

C++/test/data_structures/PriorityQueue.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+TEST_CASE("Checking max() and remove() functions return the expected value", "[PriorityQueue]"){
```
faheel: Change all the `[PriorityQueue]` tags to `[priority_queue]`.

C++/include/data_structures/PriorityQueue.cpp
```diff
@@ -0,0 +1,99 @@
...
+	2. Max-heap property i.e Parent's priority is no smaller than children's priority(can be equal).
+
+	PRIORITY IS THE HIGHEST FOR THE LARGEST VALUE (PRIORITY IS ALWAYS GREATER THAN ZER0)
+	-----------------------------------------------------------------------------------
+
+	Time complexity
+    ---------------
+    O(log N) - for insert(), Where N is number of items in priority queue
+    O(log N) - for remove() , Where N is number of items in priority queue
+    O(1) - for max() , constant amount of time
```
faheel: The time and space complexity for `insert()`, `remove()` and `max()` should be removed from here and should instead be mentioned in the description comment of the respective function. See [this file](https://github.com/faheel/Algos/blob/master/C%2B%2B/include/data_structures/disjoint_set.cpp) for an example.

C++/include/data_structures/PriorityQueue.cpp
```diff
@@ -0,0 +1,99 @@
...
+    
+    Space complexity
+    ----------------
+    O(N) - Where N is number of items in priority queue
+*/
+#include "PriorityQueue.hpp"
+#include <iostream>
+#include <vector>
+#include <stdlib.h>
```
faheel: `stdlib.h` -> `cstdlib`

C++/include/data_structures/PriorityQueue.cpp
```diff
@@ -0,0 +1,99 @@
...
+/*
+    Constructor
+*/
+PriorityQueue::PriorityQueue() {
+	v.push_back(INT_MIN); // indices of vector start at 1 
+}
+
+void PriorityQueue::heapify_up(int position) {
+	/* Function to restore the max-heap property while inserting new priorities in the priority Queue.
+	   Variable "position" represent the position at which the heap property is violated */
```
faheel: The description of each member function should be mentioned in its description comment (along with the time and space complexity, if it's an algorithmic function). See [this file](https://github.com/faheel/Algos/blob/master/C%2B%2B/include/data_structures/disjoint_set.cpp) for an example.

C++/include/data_structures/PriorityQueue.cpp
```diff
@@ -0,0 +1,99 @@
...
+		swap(v[(position/2)], v[position]);
+		position = position/2;
+	}
+}
+
+void PriorityQueue::heapify_down(int position) {
+	/* Function to restore the max-heap property while deleting priorities from the priority queue.
+	   Variable position represent the position at which the heap property is violated */
+
+	while(2*position <= static_cast<int>(v.size())-1) { // static_case<int> used to typecaste v.size() to return signed int value
```
faheel: The comment can be removed as it doesn't add any useful meaning.

C++/test/data_structures/PriorityQueue.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+	PriorityQueue pq; // creating PriorityQueue class object
```
faheel: The comment can be removed as it doesn't give any extra meaning.

C++/include/data_structures/PriorityQueue.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void PriorityQueue::heapify_up(int position) {
+	/* Function to restore the max-heap property while inserting new priorities in the priority Queue.
+	   Variable "position" represent the position at which the heap property is violated */
+
+	while (position > 1 && v[(position/2)] < v[position]) {
+
+		swap(v[(position/2)], v[position]);
+		position = position/2;
```
faheel: - The parenthesis inside `[]` can be removed
- Add a space to the left and right of `/`

C++/include/data_structures/PriorityQueue.hpp
```diff
@@ -0,0 +1,22 @@
+#ifndef PRIORITY_QUEUE
+#define PRIORITY_QUEUE
+
+#include <vector>
+
+class PriorityQueue{
+private:
+	
+	std::vector<int> v;
```
faheel: Give `v` a more meaningful name, such as `values` or `priorities`.

faheel (changes requested):
- In all the filenames, `PriorityQueue` should be changed to `priority_queue` (only the filenames; the class remains as it is)
- Please don't start a function with a blank line.
- All function parameters whose value isn't changed should be accepted as a constant reference `const type& param` instead of `type param` to prevent creating multiple copies of them.
- Use `and` and `or` instead of `&&` and `||`
- Make sure that each level of indentation is four spaces. You can convert all tabs to spaces using your text editor.

faheel (approved):

faheel:
@Kuldip397 Please update your branch with the changes from upstream and then it'll be good to merge.

alxmjo:
Closing due to inactivity.

PR #93, opened 2017-07-28
added build support for gcc < 5 with c++14

yahiafarghaly:
I am using ubuntu 14.04 LTS with gcc 4.8 . 
Your Makefile is set to build with support of c++14 and my gcc 4.8 doesn't support this build with the flag of -std=c++14. it support with this flag c++1y . so i created a conditional statement in the MakeFile where if someone has gcc that less than 5 , it will use the flag of c++1y else it will use c++14. 

more practical is to check also the minor version since -std=c++14 is supported from version 5.2 but doing comparison operations in makefile is not flexible as cmake unfortunately.

### DIFF
--- a/C++/Makefile
+++ b/C++/Makefile
@@ -1,5 +1,11 @@
 # c++ version
+GCC_MAJOR:=$(shell echo $(GCC_VERSION) | cut -d'.' -f1)
+
+ifeq ($(GCC_MAJOR),5)
 CXXFLAGS += -std=c++14
+else
+CXXFLAGS += -std=c++1y
+endif
 
 # warning flags
 CXXFLAGS += -Wall -Wextra -pedantic-errors


C++/Makefile
```diff
@@ -1,5 +1,11 @@
 # c++ version
+GCC_MAJOR:=$(shell echo $(GCC_VERSION) | cut -d'.' -f1)
+
+ifeq ($(GCC_MAJOR),5)
 CXXFLAGS += -std=c++14
```
faheel: This project will also compile on platforms where GCC may not be installed, and Clang will be used there. So you would need to add more conditions here to first check which compiler is being used, which will be the value of `CXX` environment variable. If `CXX` turns out to be `g++` then you would check the major version.

Also, you would need to change the condition to the equivalent of `$GCC_MAJOR >= 5` because there are versions 6 and 7 of GCC too (I use 7.1.1).

faheel, line 5: This project will also compile on platforms where GCC may not be installed, and Clang will be used there. So you would need to add more conditions here to first check which compiler is being used, which will be the value of `CXX` environment variable. If `CXX` turns out to be `g++` then you would check the major version.

Also, you would need to change the condition to the equivalent of `$GCC_MAJOR >= 5` because there are versions 6 and 7 of GCC too (I use 7.1.1).

faheel:
Though I suggest that you update GCC to a more recent version, as 4.8 is a little old now. Here's how you can update to the latest stable version of GCC 6 on Ubuntu 14.04:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-6 g++-6
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 60 --slave /usr/bin/g++ g++ /usr/bin/g++-6
```

faheel:
I think it would be best to let it be this way, as a lot of work would need to be done just to support a tool that is outdated and which can be updated easily. So I'm closing this for now.

In case anybody feels that it's important to support versions of GCC < 5, you may create an issue regarding this and discuss it there first.

PR #94, opened 2017-07-28
DynamicArray class

ryoung9550:
This patch features an addition of a std::vector like class called DynamicArray.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ bin
 
 ## Intermediate build files
 build
+
+
+## Swap Files
+*.swp
--- a/C++/include/data_structures/dynamic_array.hpp
+++ b/C++/include/data_structures/dynamic_array.hpp
@@ -1,0 +1,125 @@
+#include <cstdlib> // std::malloc(), std::realloc()
+
+/*
+	Brief
+	-------
+	Dynamic Array is an implementation from scratch based
+	off of the SDL std::vector
+*/
+
+template<class T>
+class DynamicArray
+{
+private:
+	unsigned arrSize;
+	unsigned numOfItems;
+	T* itemList;
+public:
+	DynamicArray();
+	~DynamicArray();
+	int push_back(T);
+	int pop_back();
+	unsigned size();
+	unsigned capacity();
+	int resize(unsigned);
+	int insert(unsigned, T);
+	T& operator[](unsigned);
+	T& at(unsigned);
+	T& front();
+	T& back();
+	bool empty();
+};
+
+template<class T>
+DynamicArray<T>::DynamicArray() : arrSize(0), numOfItems(0), itemList(nullptr) {}
+
+template<class T>
+DynamicArray<T>::~DynamicArray()
+{
+	if (itemList != nullptr) {
+		std::free(itemList);
+	}
+}
+
+template<class T>
+int DynamicArray<T>::push_back(T item)
+{
+	if (arrSize == 0 && itemList == nullptr) {
+		itemList = (T*) std::malloc(sizeof(T));
+		arrSize++;
+		itemList[0] = item;
+		numOfItems = 1;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,3 +4,7 @@ bin
 
 ## Intermediate build files
 build
+
+
+## Swap Files
+*.swp
```
faheel, line 10: Are swap files related to Vim? If so then they get deleted automatically after you save and/or exit from the editor, and so they don't need to be added to gitignore.

C++/include/data_structures/dynamic_array.hpp
```diff
@@ -0,0 +1,129 @@
...
+T& DynamicArray<T>::operator[](unsigned index) { return itemList[index]; }
+
+template<class T>
+T& DynamicArray<T>::at(unsigned index) { return itemList[index]; }
+
+template<class T>
+T& DynamicArray<T>::front()
+{
+	if (numOfItems > 0)
+		return itemList[0];
```
faheel: What about the `else` case? If `numOfItems` is 0, this will lead to an error. Similarly for `back()`.

C++/include/data_structures/dynamic_array.hpp
```diff
@@ -0,0 +1,129 @@
...
+	off of the STL std::vector
+*/
+
+template<class T>
+class DynamicArray
+{
+private:
+	unsigned arrSize;
+	unsigned numOfItems;
+	T* itemList;
```
faheel: - Use `size_t` instead of `unsigned`, as used in the STL.
- Please rename these data members so that they closely mimic those in the STL. Eg: `arrSize` can be renamed as `capacity` and `numOfItems` can be renamed as `size`. `itemList` can be renamed to just `items`.

faheel (changes requested):
This would be a good addition. Thanks!
However, there are a few changes that need to be made.
- Please use soft tabs that are 4 spaces long.
- For those function arguments whose value does not change, accept them as `const type& param` instead of `type param` to prevent multiple copies being made.
- For member functions that do not change any data members, mark them as `const`.
- Use `and` and `or` instead of `&&` and `||`.

.gitignore
ryoung9550: Swap files are temp files which vim makes when files are opened. I typically use multiple terminals with some files already open so these files show up in the git status. I'm not sure what the conventions are with .gitignore so if it doesn't fit the bill I'll remove it.

faheel: Ah, thanks for the explanation. It's OK to add it to gitignore, but please change the comment to `Swap files for Vim`

faheel (commented):
Looks good to me, but its usage needs to be tested. It would be great if you could also add a unit test for DynamicArray. The instructions can be found in [UNIT_TESTS.md](https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md).

C++/test/data_structures/dynamic_array.test.cpp
```diff
@@ -0,0 +1,15 @@
+#define DYNAMIC_ARRAY
+#define CATCH_CONFIG_MAIN
+
+#include "third_party/catch.hpp"
+#include "source/path_to/source_file.cpp"
+
+// Add your test cases here:
+#ifndef DYNAMIC_ARRAY
+
+
```
faheel, line 10: Looks like you forgot to add the tests :)

In case you aren't quite sure about what kind of tests to add or how to add them, you can take a look at the [unit test for DisjointSet](https://github.com/faheel/Algos/blob/master/C%2B%2B/test/data_structures/disjoint_set.test.cpp) for some inspiration.

faheel:
Quite a few changes have been made to the project, so please pull the changes from upstream to update your fork.

alxmjo:
Closing due to inactivity.

Issue #79, opened 2017-07-07
Median in a stream of integers


SKAUL05:
I would like to add an algorithm that calculates the median in a stream of running integers.
Time Complexity:O(N^2)

faheel:
Can you please explain what a "stream of running integers" means? I take it that it's different from simply a container (`array`, `vector`, `set` etc.) of integers.

SKAUL05:
We would be given a number containing the number of integers to follow. Taking input of every integer we have to print the median of current set of integers.
e.g when reading 1st element say 5-> median=5
when reading 2nd element say 25-> median =25
when reading 3rd element say 12-> median =12.
like so on... 
We will take a vector to push integers sequentially.

faheel:
Thanks for the clarification! It would be nice to have it in repo.
But I think the time complexity should be **_O_**(N * log(N)). If you use a `std::set` or a `std::multiset`, each insertion will be an **_O_**(log(N)) operation. Then finding the median will be an **_O_**(1) operation as you would simply return the element at index `set.size() / 2` (or the average of `set.size() / 2` and `set.size() / 2 + 1` if the size is even).
So for a total of `N` numbers, the overall time complexity will be:
log(1) + log(2) + log(3) + ... + log(N)
= log(N!)
= **_O_**(N * log(N)), or to be more accurate, **_Θ_**(N * log(N)).

SKAUL05:
Yeah it is O(N * log(N)) but i am using a vector and then inserting new element by finding upper_bound as the stream needs to be in increasing order.

faheel:
Using a set or multiset would be simpler and more efficient. Then why use a vector?

SKAUL05:
There is no specific point to use vectors but i'm much more familiar with vectors and since both have same time complexities i preferred vectors.

faheel:
> [...] since both have same time complexities [...]

Only if the stream is in increasing order. Using a multiset, this constraint will be eliminated. And having less constraints is more preferable :)

I would suggest reading up on [`std::multiset`](http://en.cppreference.com/w/cpp/container/multiset) and using it to implement the algorithm. [`std::advance`](http://en.cppreference.com/w/cpp/iterator/advance) will also be of help in accessing element by index in a multiset.

SKAUL05:
The stream is not in increasing order. The stream can be in any order we have to arrange it in increaing order so as to find the median. Only assumption is that there are no duplicates.

faheel:
[...]

PR #95, opened 2017-08-02
Added median of a stream of integers

SKAUL05:
Please look at my program and merge it.

### DIFF
--- a/C++/source/number_theory/Median_In_a_Stream.cpp
+++ b/C++/source/number_theory/Median_In_a_Stream.cpp
@@ -1,0 +1,77 @@
+/*
+    Median in a stream of integers(running integers)
+    ------------------------------------------------
+    An algorithm to find the median of elements read so far from
+    a data stream in a efficient way.
+
+    Assumptions: There are no duplicates.
+
+    Time Complexity
+    ---------------
+    O(NlogN), where N is the number of elements.
+
+    Space Complexity
+    ----------------
+    O(N^2)
+
+    Auxilliary Space
+    ----------------
+    O(N)
+
+*/
+//Explanation link :https://www.youtube.com/watch?v=VmogG01IjYc
+#include<iostream>
+#include<algorithm>
+#include<set>
+#include<math.h>
+#include<iterator>
+#include<queue>
+
+using namespace std;
+
+typedef long long ll;
+priority_queue <ll> great;
+priority_queue<ll,vector <ll>,greater<ll> > shorter;
+
+void add(ll num)
+{
+	if(!great.empty() and num < great.top())
+	{
+		great.push(num);
+		if (great.size() - shorter.size() == 1)
+		{
+			shorter.push(great.top());
+			great.pop();
+		}
+	}
+	else
+	{
+		shorter.push(num);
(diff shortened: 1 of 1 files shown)


C++/source/number_theory/Median_In_a_Stream.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include<set>
+#include<math.h>
+#include<iterator>
+#include<queue>
+
+using namespace std;
+
+typedef long long ll;
+priority_queue <ll> great;
+priority_queue<ll,vector <ll>,greater<ll> > shorter;
```
faheel, line 34: Make these queues local to `main()`

C++/source/number_theory/Median_In_a_Stream.cpp
```diff
@@ -0,0 +1,77 @@
...
+		return (shorter.top() + great.top()) / 2.0;
+	return shorter.top();
+}
+int main()
+{
+	ios::sync_with_stdio(false);
+	cin.tie(NULL);
+	ll t;
+	cin>>t;
+	while(t--)
```
faheel: Give a better name for `t`, such as `num_elements`. Also, print a prompt for accepting the input from the user. Eg.
> Enter the number of elements : 

On another note, it would be better to not ask for the number of integers, but instead have a way of determining when to stop reading them. This could be done by using some string (eg. `---`) as the end marker, but for this the input should be taken as a string, checked if it is not the end marker, and then converted to an integer to calculate the median.

C++/source/number_theory/Median_In_a_Stream.cpp
```diff
@@ -0,0 +1,77 @@
...
+	{
+		ll x;
+		cin>>x;
+		add(x);
```
faheel: Give a more appropriate name to `x` such as `num`. Also, declaring it outside the loop will be more efficient.

faheel (changes requested):

SKAUL05:
Please review the changes made

C++/source/number_theory/Median_In_a_Stream.cpp
faheel: These queues should be declared inside `main()` and passed to whichever function needs them by reference.

C++/source/number_theory/Median_In_a_Stream.cpp
```diff
@@ -0,0 +1,78 @@
...
+	if (shorter.size() == great.size())
+		return (shorter.top() + great.top()) / 2.0;
+	return shorter.top();
+}
+int main()
+{
+    ios::sync_with_stdio(false);
+	cin.tie(NULL);
+	ll num_elements;
+	cin>>num_elements;
```
faheel, line 69: Please make sure that your code is properly indented, using 4 spaces for a level of indentation. You can easily convert all tabs to spaces using your text editor / IDE.

faheel (changes requested):

faheel:
It would be great if you can add unit-test for your code too.

faheel:
Closes #79

alxmjo:
Closing due to inactivity.

PR #50, opened 2017-04-30
Create Tower_of_Hanoi.cpp

Satyamjay:
Added tower of hanoi algorithm in recursion folder

PR #102, opened 2017-08-26
Tower of hanoi

ShumWengSang:
C++ algorithm to solving the classic Tower of Hanoi problem by using recursion.


faheel (changes requested):
Using a class for implementing a solution to the Tower of Hanoi problem is a bit "overkill" in my opinion. Keeping things simple is always the best approach ([KISS principle](https://en.wikipedia.org/wiki/KISS_principle)).

Please take a look at #50 to get an idea of how simple the implementation can be made. Thanks for your understanding :)

PR #104, opened 2017-09-07
Added 

anand434:
now it can compute GCD for any number a, b or b, a

### DIFF
--- a/C++/source/number_theory/gcd.cpp
+++ b/C++/source/number_theory/gcd.cpp
@@ -19,11 +19,15 @@
 #include <iostream>
 using namespace std;
 
+// Function to return maximum/minimum of a and b
+int max(int a, int b) { a>b ? a : b; }
+int min(int a, int b) { a<b ? a : b; }
+
 // Function to return gcd of a and b
 int gcd(int a, int b) {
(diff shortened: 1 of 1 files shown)


C++/source/number_theory/gcd.cpp
```diff
@@ -19,11 +19,15 @@
...
+int min(int a, int b) { a<b ? a : b; }
+
 // Function to return gcd of a and b
 int gcd(int a, int b) {
-    if (a == 0)
-        return b;
-    return gcd(b%a, a);
+    if (min(a, b) == 0)
+        return max(a, b);
+    return gcd(max(a, b) % min(a, b), min(a, b));
```
faheel, line 30: The existing code already takes care of what you're trying to add.

Consider this example case:
`gcd(152, 74)` :arrow_right: `gcd(74, 152)` :arrow_right: `gcd(4, 74)` :arrow_right: `gcd(2, 4)` :arrow_right: `gcd(0, 2)` :arrow_right: `2`

Whether you call `gcd(larger_num, smaller_num)` or `gcd(smaller_num, larger_num)`, the result will be the same.

PR #112, opened 2017-10-03
Add linear search with test

### LINKED ISSUE #111: Linear search
Hey, I just came across this repository via Hacktoberfest.
It's pretty interesting and I would like to contribute! I saw that there's a very elementary algorithm missing in "Searching" - Linear Search. I was wondering if I could start off by authoring that?

faraazahmad:
Closes #111

### DIFF
--- a/C++/source/searching/linear_search.cpp
+++ b/C++/source/searching/linear_search.cpp
@@ -18,14 +18,14 @@
 
 using namespace std;
 
-int linearSearch(const int value, const vector<int> &vect) {
+size_t linear_search(const int value, const vector<int> &vect) {
 
     // Search all the elements in the array
-    for(i = 0; i < vect.length(); i++) {
+    for(unsigned i = 0; i < vect.size(); i++) {
         // If the element at index i matches the target value
         // return the position i
         if(vect[i] == value) {
-            return i;
+            return i + 1;
         }
     }
 
@@ -33,6 +33,7 @@ int linearSearch(const int value, const vector<int> &vect) {
     return -1;
 }
 
+#ifndef LINEAR_SEARCH_TEST
 int main() {
     int value;
     cout << "Enter the value to search for: ";
@@ -47,12 +48,13 @@ int main() {
     for(int& val: inputVect)
         cin >> val;
 
-    int index = linearSearch(value, inputVect);
+    int index = linear_search(value, inputVect);
     cout << "\n";
     if(index != -1)
-        cout << "Found " << value << " at position " << (index + 1) << "\n";
+        cout << "Found " << value << " at position " << index << "\n";
     else
-        cout << "Either " << value << " is not present among the input values\n"
+        cout << value << " is not present among the input values\n";
 
     return 0;
 }
+#endif
--- a/C++/source/searching/linear_search.cpp
+++ b/C++/source/searching/linear_search.cpp
@@ -18,10 +18,11 @@
 
 using namespace std;
 
-size_t linear_search(const int value, const vector<int> &vect) {
+template <typename T>
+size_t linear_search(const T value, const vector<T> &vect) {
 
     // Search all the elements in the array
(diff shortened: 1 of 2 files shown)


C++/source/searching/linear_search.cpp
```diff
@@ -0,0 +1,58 @@
...
+    --------------
+    O(1)
+*/
+
+#include <iostream>
+#include <vector>
+
+using namespace std;
+
+int linearSearch(const int value, const vector<int> &vect) {
```
faheel: * Return type should be `size_t`
* Function's name should be `linear_search`. See [function naming rules](https://github.com/faheel/Algos/blob/master/C%2B%2B/CODING_GUIDELINES.md#function-names).

C++/source/searching/linear_search.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    // Search all the elements in the array
+    for(i = 0; i < vect.length(); i++) {
```
faheel: * `i` should be declared as a `size_t`
* `length()` does not exist for a `vector`. The right function to use is [`size()`](http://en.cppreference.com/w/cpp/container/vector/size).

C++/source/searching/linear_search.cpp
```diff
@@ -0,0 +1,58 @@
...
+    cout << "Enter " << size << " integers: ";
+    for(int& val: inputVect)
+        cin >> val;
+
+    int index = linearSearch(value, inputVect);
+    cout << "\n";
+    if(index != -1)
+        cout << "Found " << value << " at position " << (index + 1) << "\n";
+    else
+        cout << "Either " << value << " is not present among the input values\n"
```
faheel: "Either" needs to be removed (under the `else` block)

C++/source/searching/linear_search.cpp
faheel: It would be better if this function is generic, i.e. it works not only for integers but for all types.

See http://www.cplusplus.com/doc/oldtutorial/templates/ for an introduction on templates in C++.

C++/test/searching/linear_search.test.cpp
```diff
@@ -0,0 +1,17 @@
+#define LINEAR_SEARCH_TEST
+#define CATCH_CONFIG_MAIN
+
+#include "third_party/catch.hpp"
+#include "source/searching/linear_search.cpp"
+
+TEST_CASE("Base cases", "[linear search]") {
+    REQUIRE(linear_search(1, vector<int>({1})) == 1);
+    REQUIRE(linear_search(1, vector<int>({0})) == -1);
```
faheel: Also check that the search returns `-1` when the vector is empty (`vector<int>()`).

C++/test/searching/linear_search.test.cpp
```diff
@@ -0,0 +1,17 @@
...
+}
+
+TEST_CASE("Normal cases", "[linear search]") {
+    REQUIRE(linear_search(5, vector<int>({0, -5, 25, 8, 5, 2, 6, 3})) == 5);
+    REQUIRE(linear_search(-15, vector<int>({9, -15, 0, 3, 5})) == 2);
+    REQUIRE(linear_search(0, vector<int>({1, 29, 6, 8, 5, 2, 0})) == 7);
+    REQUIRE(linear_search(-0, vector<int>({0, -5, 25, 8, 5, 2, 6, 3})) == 1);
+}
```
faheel: Missing
```C++
#undef LINEAR_SEARCH_TEST
```
at the end. See [Unit tests for C++](https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md).

faheel (changes requested):

faheel:
@faraazahmad Thank you for your time and contribution! :stuck_out_tongue_winking_eye:

PR #118, opened 2017-10-10
Add Euler's Sieve

lonelam:
Hey, Since we have a sieve function here why don't we upgrade it into a more efficient one?
I just copy the eratos sieve and modify it a bit.
Euler's Sieve has less time complexity and use a bit more space (you need to save prime numbers) than eratos sieve.


alxmjo:
@lonelam Thanks for contributing. We've made a few of changes to the repository since you opened this PR. You can read about them in the [documentation](https://github.com/faheel/Algos/blob/master/README.md) – the main one is that we now implement each algorithm and data structure as a header file (in the `include` directory) and then test them via .cpp files (in the `test` directory).

I'll leave this open for a couple weeks in case you'd like to update your PR so that it fits the new standards.

lonelam:
closed. 
thanks.

PR #120, opened 2017-10-15
Permutations of a String

MeLlamoArroz:
Hi!

The algorithm prints all the permutations of a string, prints duplicates only if the string have duplicates characters.

### DIFF
--- a/C++/source/number_theory/trial_division.cpp
+++ b/C++/source/number_theory/trial_division.cpp
@@ -1,0 +1,63 @@
+/*
+	Trial Division
+	_______________
+	Check if a number is prime or not, given the value of N.
+	
+	Checks if a prime in the range of 2 to sqrt(N) divides N.
+	If it does, then N is not a Prime, otherwise, N is a Prime.
+
+	Time Complexity
+	________________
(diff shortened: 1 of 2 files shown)


C++/source/number_theory/trial_division.cpp
```diff
@@ -0,0 +1,63 @@
+/*
+	Trial Division
+	_______________
+	Check if a number is prime or not, given the value of N.
```
faheel, line 4: I think you added this file by mistake.

C++/source/string/permutations.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include<iostream>
+#include<string.h>
+
+using namespace std;
+
+void permutations(string& a, int first, int last)
+{
+   	if(first==last)
+		cout<<a<<endl;
```
faheel, line 28: Instead of printing all the permutations, how about passing a `std::set` by reference, and inserting permutations in the set. This will also eliminate duplicates as a `set` only stores a value once.

Also, this will make it easier to check the algorithm by writing tests that will check whether after the function is executed, the set contains the expected permutated strings for a given input string.

alxmjo:
Looks like this hasn't been updated in a few months. Closing.

PR #121, opened 2017-10-15
Added Linear Search

MeLlamoArroz:
Done!!


faheel:
Linear search has already been added

PR #124, opened 2017-10-18
8puzzle added (by BFS, DFS and A* algorithm)

### LINKED ISSUE #122: 8-puzzle problem using BFS, DFS (stack and recursion) and A* algorithm
I would like to work on 8 puzzle problem

nikki1016:
8 puzzle implementation using BFS,DFS and A* algorithm is added. IT is implemented as per the algorithm described in the book on Artificial-Intelligence by Stuart-Russell-Peter-Norvig-Artificial-Intelligence_-A-Modern-Approach

Closes #122


faheel:
@nikki1016 Thank you for contributing, and sorry for the late response.

It would be great if you could make a different source file for each algorithm that you've added, and place them all under `C++/source/ai/` as `8_puzzle_bfs.cpp`, `8_puzzle_dfs.cpp` and `8_puzzle_a_star.cpp`.

faheel (changes requested):
I see a lot of code that is being repeated in the 3 files. Please separate out the common code and add it in a header file named `8_puzzle.hpp` in `C++/include/ai/`, which you can include in each of the 3 programs using:
```C++
#include "ai/8_puzzle.hpp"
```

alxmjo:
Looks like this hasn't been updated in a while. Closing.

PR #125, opened 2017-10-19
Finding the absolute value of an integer without branching.

ahmedkrmn:
A simple yet tricky algorithm based on the bit manipulation of the given integer, the algorithm returns the absolute value of the integer without the usage of the built in function abs() or any branching techniques.


alxmjo:
Thanks for contributing, @ahmedkrmn. We've made a few of changes to the repository since you opened this PR. You can read about them in the [documentation](https://github.com/faheel/Algos/blob/master/README.md) – the main one is that we now implement each algorithm and data structure as a header file (in the `include` directory) and then test them via .cpp files (in the `test` directory).

I'll leave this open for a couple weeks in case you'd like to update your PR so that it fits the new standards.

PR #130, opened 2017-10-24
Added Edit Distance Algorithm in String Algorithms

### LINKED ISSUE #97: Edit distance algorithm
I would like to implement edit distance algorithm for strings

mayanktaluja:
Closes #97.

### DIFF
--- a/C++/source/string/edit_distance.cpp
+++ b/C++/source/string/edit_distance.cpp
@@ -1,0 +1,45 @@
+/*
+
+Edit Distance Algorithm :-
+--------------------------------
+Description: Find the minimum number of operations required to convert one string into another.
+
+Time Complexity :- O(m * n)
+Space Complexity :- O(m * n)
+                    Where m = length of string1 and n = string2
+
+*/
+
+
+
+#include<bits/stdc++.h>
+
+using namespace std;
+ 
+int min(int x, int y, int z) {
+   return min(min(x, y), z);
+}
+ 
+ int editDist(string str1 , string str2 , int m ,int n) {
+    if (m == 0) return n;
+ 
+    if (n == 0) return m;
+ 
+    if (str1[m-1] == str2[n-1])
+        return editDist(str1, str2, m-1, n-1);
+ 
+    return 1 + min ( editDist(str1,  str2, m, n-1),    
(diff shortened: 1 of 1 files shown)


C++/source/string/edit_distance.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+Time Complexity :- O(m * n)
+Space Complexity :- O(m * n)
+                    Where m = length of string1 and n = string2
+
+*/
+
+
+
+#include<bits/stdc++.h>
```
faheel, line 15: It's good practice to only `#include` those headers that are required.

Also, this header is not present on OS X, as so the code does not compile on it.

C++/source/string/edit_distance.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+using namespace std;
+ 
+int min(int x, int y, int z) {
+   return min(min(x, y), z);
+}
```
faheel, line 21: No need for this function. You can instead use `min({some_val, some_other_val, yet_another_val})` in your code.

C++/source/string/edit_distance.cpp
```diff
@@ -0,0 +1,45 @@
...
+    return 1 + min ( editDist(str1,  str2, m, n-1),    
+                     editDist(str1,  str2, m-1, n),   
+                     editDist(str1,  str2, m-1, n-1)
+                   );
+}
+ 
+int main()
+{
+    string str1 = "sunday";
+    string str2 = "saturday";
```
faheel, line 40: To make it more dynamic, please take the input for `str1` and `str2` from the user.

C++/source/string/edit_distance.cpp
```diff
@@ -0,0 +1,45 @@
+/*
+
+Edit Distance Algorithm :-
+--------------------------------
+Description: Find the minimum number of operations required to convert one string into another.
+
+Time Complexity :- O(m * n)
+Space Complexity :- O(m * n)
+                    Where m = length of string1 and n = string2
```
faheel, line 9: The space complexity is O(1).

faheel (changes requested):

alxmjo:
Looks like this hasn't been updated in a few months. Closing.

PR #131, opened 2017-10-26
Explained #ifndef

aditya-ml:
#ifndef has been explained for reference


faheel:
It's not really necessary, so I'm closing the PR now. Thank you for deciding to contribute though! If you want you can add an implementation of an algorithm or data structure, or improve an existing one.

PR #137, opened 2017-10-29
Add default values for parameters used in sorting functions

### LINKED ISSUE #136: Add default values to parameters for sorting functions
Almost all of the sorting functions in this repository have the parameters `order` and `to_show_state` to which the default values `1` and `false` need to be assigned respectively.

Naman-sopho:
Fixes #136

### DIFF
--- a/C++/include/sorting/utils.hpp
+++ b/C++/include/sorting/utils.hpp
@@ -12,6 +12,10 @@
 
 using namespace std;
 
+//default values
+const int order = 1;
+const bool to_show_state = false;
+
 /*
     Input utils
     -----------


C++/include/sorting/utils.hpp
```diff
@@ -12,6 +12,10 @@
 
 using namespace std;
 
+//default values
+const int order = 1;
+const bool to_show_state = false;
```
faheel, line 17: Adding these here won't solve the issue. You need to set default values for the function parameters in the definition of each sorting function that you find in various files under `source/sorting/`.

faheel (changes requested):

C++/include/sorting/utils.hpp
namtiw: @faheel I am on it.

faheel:
#136 is now closed.

PR #139, opened 2017-10-31
Add default values for order and to_show_state

### LINKED ISSUE #136: Add default values to parameters for sorting functions

alxmjo:
Adds default values for order and to_show_state in function declarations for each sorting function, with the exception of radix_sort, which takes those arguments by reference. Fixes #136.

### DIFF
--- a/C++/source/sorting/bubble_sort.cpp
+++ b/C++/source/sorting/bubble_sort.cpp
@@ -21,7 +21,7 @@
 
 using namespace std;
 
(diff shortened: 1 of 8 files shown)


faheel:
Thanks for helping out on this issue @alxmjo!

alxmjo:
My pleasure. 🙂

PR #140, opened 2017-11-02
added fractional knapsack algorithm

youteekarsh:
Hello Sir, Please see my code for the fractional knapsack problem.

If you find it acceptable then please add it or otherwise  guide me about this algorithm.

### DIFF
--- a/C++/source/knapsack_problem/fractional_knapsack.cpp
+++ b/C++/source/knapsack_problem/fractional_knapsack.cpp
@@ -1,0 +1,111 @@
+/*
+    Fractional Knapsack Problem
+    ---------------------------
+    This is a Greedy Algorithm which calculates the maximum
+    profit which can be earned from a given set of unsorted
+    objects with associated weight and profit.
+    Class implementation using Selection sort
+    for more simplicity.
+
+    Time Complexity
+    ---------------
+    O(n^2), where n is the number of objects
+
+    Space Complexity
+    ----------------
+    O(1)
+*/
+
+
+#include <iostream>
+
+using namespace std;
+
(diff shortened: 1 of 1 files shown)


C++/source/knapsack_problem/fractional_knapsack.cpp
```diff
@@ -0,0 +1,111 @@
...
+    O(1)
+*/
+
+
+#include <iostream>
+
+using namespace std;
+
+class object {
+    public:
```
faheel, line 25: Since everything is public, you should change `Object` to a `struct` instead.

C++/source/knapsack_problem/fractional_knapsack.cpp
```diff
@@ -0,0 +1,111 @@
...
+            cout<<"profit:";
+            cin>>profit;
+        }
+        void calc_ratio ()
+        {
+            p_w_ratio = profit/weight;
+            cout<<"ratio:"<<p_w_ratio;
+            cout<<endl;
+        }
+}obj[50];
```
faheel, line 43: Don't create a global object. Declare it in `main()` instead.

C++/source/knapsack_problem/fractional_knapsack.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+void swap_data(float *x, float *y)
+{
+float temp;
+temp=*x;
```
faheel, line 48: There's no need for this function. You can instead use the `swap` function from the `algorithm` header file.

faheel (changes requested):

PR #150, opened 2018-01-18
Jump Search and exponential search

asas2016SEC:


C++/source/searching/jump_search.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return -1;
+}
+ 
+// Driver program to test function
+int main()
+{
+    int arr[] = { 0, 1, 1, 2, 3, 5, 8, 13, 21,
+                 34, 55, 89, 144, 233, 377, 610 };
+    int x = 55;
+    int n = sizeof(arr) / sizeof(arr[0]);
```
faheel, line 46: You need to take input from the user. Take a look at the [code for linear search](https://github.com/faheel/Algos/blob/master/C%2B%2B/source/searching/linear_search.cpp), from where you can copy the code for taking user input.

C++/source/searching/Exponential_search.cpp
```diff
@@ -0,0 +1,66 @@
...
+    return -1;
+}
+ 
+// Driver code
+int main(void)
+{
+   int arr[] = {2, 3, 4, 10, 40};
+   int n = sizeof(arr)/ sizeof(arr[0]);
+   int x = 10;
+   int result = exponentialSearch(arr, n, x);
```
faheel, line 61: You need to take input from the user. Take a look at the [code for linear search](https://github.com/faheel/Algos/blob/master/C%2B%2B/source/searching/linear_search.cpp), from where you can copy the code for taking user input.

C++/source/searching/Exponential_search.cpp
```diff
@@ -0,0 +1,66 @@
+// C++ program to find an element x in a
+// sorted array using Exponential search.
+#include <bits/stdc++.h>
```
faheel, line 3: This does not work for all compilers. Include only those headers that are required.

C++/source/searching/jump_search.cpp
```diff
@@ -0,0 +1,54 @@
+// C++ program to implement Jump Search
+ 
+#include <bits/stdc++.h>
```
faheel, line 3: This does not work for all compilers. Include only those headers that are required.

C++/source/searching/Exponential_search.cpp
```diff
@@ -0,0 +1,66 @@
+// C++ program to find an element x in a
+// sorted array using Exponential search.
```
faheel, line 2: Please read the [contribution guidelines](https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md) to see how to format this comment.

C++/source/searching/jump_search.cpp
```diff
@@ -0,0 +1,54 @@
+// C++ program to implement Jump Search
```
faheel, line 1: Please read the [contribution guidelines](https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md) to see how to format this comment.

C++/source/searching/Exponential_search.cpp
```diff
@@ -0,0 +1,66 @@
...
+        i = i*2;
+ 
+    //  Call binary search for the found range.
+    return binarySearch(arr, i/2, min(i, n), x);
+}
+ 
+// A recursive binary search function. It returns
+// location of x in  given array arr[l..r] is
+// present, otherwise -1
+int binarySearch(int arr[], int l, int r, int x)
```
faheel, line 29: Binary search [already exists](https://github.com/faheel/Algos/blob/master/C%2B%2B/source/searching/binary_search.cpp). It would be better if you found a way of reusing it here, rather than reimplementing it.

faheel (changes requested):
* Please read the [C++ coding guidelines](https://github.com/faheel/Algos/blob/master/C++/CODING_GUIDELINES.md) and modify the function, variables and file names as necessary.
* Use C++'s `cin` and `cout` instead of C's `scanf` and `printf`.
* Add unit tests for each of the source files. You can take help from [existing tests for searching algorithms](https://github.com/faheel/Algos/tree/master/C%2B%2B/test/searching).

PR #152, opened 2018-01-26
Implementation of RSA algorithm

dfotis:
I have implemented the RSA cryptographic algorithm for small keys. In the future I will use the BigInt (Arbitrary-sized integer class for C++) in order to make it work with large keys.

- [x] Closes an issue: close # 145


alxmjo:
Looks like this has been inactive for a few months. Closing.

PR #155, opened 2018-01-27
Added Bucket Sort

### LINKED ISSUE #63: Bucket sort

prateekkalra:
- [x] Have read the [contribution guidelines][contrib-guidelines]
- [x] Closes an issue: close #63  
- [ ] Includes unit tests (or unit tests have already been added)

### DIFF
--- a/C++/source/sorting/bucket-sort.cpp
+++ b/C++/source/sorting/bucket-sort.cpp
@@ -1,0 +1,46 @@
+/*
+    Bucket sort
+    -----------
+    A sorting algorithm that works by distributing the elements of an array into a number of buckets.
+    Each bucket is then sorted individually, either using a different sorting algorithm, or by recursively
+    applying the bucket sorting algorithm.
+    Time complexity
+    ---------------
+    O(N^2), where N is the number of elements.
+    Space complexity
+    ----------------
+    O(N), where N is the number of elements.
+*/
+
+#include <iostream>
+#include <algorithm>
+#include <vector>
+using namespace std;
+ 
+void bucketSort(float arr[], int n)
+{
+    vector<float> b[n];
+    for (int i=0; i<n; i++)
+    {
+       int bi = n*arr[i]; // Index in bucket
+       b[bi].push_back(arr[i]);
+    }
+    for (int i=0; i<n; i++)
+       sort(b[i].begin(), b[i].end());
+    int index = 0;
+    for (int i = 0; i < n; i++)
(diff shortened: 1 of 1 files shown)


C++/source/sorting/bucket-sort.cpp
```diff
@@ -0,0 +1,46 @@
...
+    int index = 0;
+    for (int i = 0; i < n; i++)
+        for (int j = 0; j < b[i].size(); j++)
+          arr[index++] = b[i][j];
+}
+ 
+int main()
+{
+    float arr[] = {0.897, 0.565, 0.656, 0.1234, 0.665, 0.3434};
+    int n = sizeof(arr)/sizeof(arr[0]);
```
faheel: As mentioned in #163, instead of having a `main()` that only runs a single test case, you can move the rest of the code into a `.hpp` file within `include/algorithm/sorting/bucket_sort.hpp`, and add a few test cases for it under `test/algorithm/sorting/bucket_sort.test.cpp`

C++/source/sorting/bucket-sort.cpp
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+#include <iostream>
+#include <algorithm>
+#include <vector>
+using namespace std;
+ 
+void bucketSort(float arr[], int n)
+{
+    vector<float> b[n];
```
faheel, line 22: Use more meaningful names than just `n` or `b`, like `size` and `buckets`.

C++/source/sorting/bucket-sort.cpp
faheel, line 39: As mentioned in #163, instead of having a `main()` that only runs a single test case, you can move the rest of the code into a `.hpp` file within `include/algorithm/sorting/bucket_sort.hpp`, and add a few test cases for it under `test/algorithm/sorting/bucket_sort.test.cpp`

C++/source/sorting/bucket-sort.cpp
faheel, line 22: It's better to use vectors (`std::vector`) instead of C-style arrays. This would also eliminate the need of another argument for the size of the array.

faheel (changes requested):
This is a very specific version of bucket sort that only works for a float array with elements with values between 0 and 1. A more general templated version would be very useful, one that can sort arrays of integers, strings, etc.

Issue #129, opened 2017-10-22
Testing sorting algorithms


alxmjo:
I'm trying to write tests for the sorting algorithms. I'm new to Catch (and unit testing in general), but based on my research I would expect this code (along with the rest of the setup you describe in `UNIT_TESTS.md`) to work:

```
TEST_CASE("Normal cases", "[bubble_sort]") {
	vector<int> unsorted1 {4, 3, 5, 2, 1};
	bubble_sort(unsorted1, 1, false);
	vector<int> sorted1 {1, 2, 3, 4, 5};
    REQUIRE(unsorted1 == sorted1);
}
```
Unfortunately when I run `make test` from the `C++` directory I receive many, many lines in the terminal that look like this:

```
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x74726f70)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x534e4573)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x00000000)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x65526576)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x74617453)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous GCDA file: corrupt arc tag (0x00000000)
profiling: /Users/alex/Programming/Algos/C++/build/sorting/bubble_sort.test.gcda: cannot merge previous run count: corrupt object tag (0x00000000)
```

If, however, I include a "dummy" test à la `REQUIRE(1 == 1)` then it works fine.

Any idea what I'm doing wrong here?

Andreasrn:
Try setting this options

`GCC_GENERATE_TEST_COVERAGE_FILES = NO;`
`GCC_INSTRUMENT_PROGRAM_FLOW_ARCS = NO;  `

Source: http://imnotyourson.com/xcode-7-unit-testing-super-slow/

alxmjo:
[...]

Issue #163, opened 2018-02-20
Reorganise C++ code into header files


faheel:
In order to maximise code reuse, it would be better to organise all C++ code into header files. This would allow one to include code from one file, say for `quicksort()`, and use in another file without reimplementing it again. This is not possible for most of the code now since it exists in `.cpp` files having a `main()` function, and so cannot be included in other files as the `main()` gets included as well.

This will also encourage writing unit tests for the headers because due to the absence of a `main()` function, the only other way to check that the code compiles and runs properly is to write a unit test for with a few cases and check that the result is as expected.

faheel:
This task is currently ongoing on the [`reorganise` branch](https://github.com/faheel/Algos/tree/reorganise).

ankur54:
I'm interested to work in this issue!! Thus, please explain it!!

alxmjo:
Hey @ankur54, thanks for your interest. 

Basically, all the algorithms will move into `include/algorithm/<type of algorithm>/<filename>.hpp` (if it's a data structure, it will move into `include/data_structure/<filename>.hpp`). And then corresponding unit tests which actually execute that algorithm will be either moved to or created in the corresponding `test` folder, i.e., `test/algorithm/<type of algorithm>/<filename>.cpp`. Each header file should have one unit test file and vice versa.

Let me know if you have any questions.

ankur54:
Ok so, I need to create both the header (.hpp) file and the unit test file and include both of them in the include folder. Is that right?

ankur54:
What about Random Algorithm and data structures?

alxmjo:
@ankur54 Header goes in `/include` and unit tests (as `.cpp` file) go in `/test`. #202 will show you how it all ends up. Same goes for data structures, etc.

PR #157, opened 2018-01-29
Add binary_search unit test

AssaSch:
added unit test for binary_search algorithm.

- [x] Have read the [contribution guidelines][contrib-guidelines]
- [ ] Closes an issue: close # 
- [x] Includes unit tests (or unit tests have already been added)

[contrib-guidelines]: https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md

PR #168, opened 2018-04-25
Add sorting tests

camilne:
Added tests for the sorting algorithms. See #129, which hasn't been updated for 5 months.

### DIFF
--- a/C++/README.md
+++ b/C++/README.md
@@ -29,15 +29,15 @@ Implementation of well-known (and some rare) algorithms, in C++.
   * [Linear search](source/searching/linear_search.cpp) :white_check_mark:
   * [Ternary search](source/searching/ternary_search.cpp) :white_check_mark:
 * Sorting
-  * [Bubble sort](source/sorting/bubble_sort.cpp)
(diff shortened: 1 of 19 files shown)


camilne:
Note that the radix sort test fails, but this is because radix sort is broken (a couple of parameters are being passed incorrectly within the function).

faheel:
@camilne Thanks for contributing! As I mentioned in #163 I am reorganising most of the code to facilitate maximum code reuse. Also, from the discussion for sorting tests in #157, it'll be good that tests reuse as much code as possible too. So I suggest that you wait until #163 is closed and then we'll discuss this PR. Again, thanks for deciding to contribute but I suggest commenting on an existing issue or creating one yourself before creating a PR so that you can be pointed in the right direction as to what next steps you should take.

faheel:
And yes, thanks for pointing out that radix sort is broken. Please create a new issue for it so that others can look into it and maybe we get a fix for it soon.

alxmjo:
Sorting tests added in #251.

PR #201, opened 2018-09-12
Migrate greatest common denominator algorithm

alxmjo:
Migrate greatest common denominator algorithm to header file (as described in #163).


alxmjo:
`greatest_common_denominator` is a pretty long filename. Also considered `euclidean_gcd`. That way its sister algorithm could be `extended_euclidean_gcd`. I'm open to input. 

Edit: On second look, that other algorithm is doing something different. So maybe `greatest_common_denominator.hpp` makes the most sense for this one, and the other one can be `extended_euclidean.hpp`.

faheel:
I haven't heard/read about greatest common "denominator" before. "Divisor" is the commonly used term for a number that divides another number. So we should stick to it.

From Wikipedia, other common terms used for GCD are:
> [...] greatest common factor (gcf), highest common factor (hcf), greatest common measure (gcm), or highest common divisor.

alxmjo:
Works for me, and shorter too. Will update and commit right now.

faheel (approved):

PR #222, opened 2018-10-12
Migrate Kadane's Algorithm

alxmjo:
Create Kadane's Algorithm as header file.
Create unit test file.
Delete old Kadane's .cpp file.
Update README.

### DIFF
--- a/C++/include/algorithm/dynamic_programming/kadane.hpp
+++ b/C++/include/algorithm/dynamic_programming/kadane.hpp
@@ -24,7 +24,7 @@ using std::make_tuple;
 using std::vector;
 
 /*
-    maximumSubarray
+    maximum_subarray
     ---------------
     Takes an array of integers as an argument and computes the maximum sum that
     can be computed from any (contiguous) subarray of that array. Uses Kadane's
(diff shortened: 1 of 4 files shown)


C++/include/algorithm/dynamic_programming/kadane.hpp
```diff
@@ -1,23 +1,41 @@
...
+    Algorithm, which employs dynamic programming.
+
+    Return value
+    ------------
+    tuple<int, size_t, size_t>, in which int is the maximum subarray value and
+    size_t to size_t represents the indices of the passed subarray that sum to
+    that value.
+*/
 
 tuple<int, size_t, size_t> maximumSubarray(const vector<int> &values)
```
faheel: The function should be renamed to `maximum_subarray`. Also, we can have an alias for it named `kadane` like so:

```cpp
int f1(int x, int y) {
    // actual definition
    return x + y;
}

int f2(int x, int y) {
    // simple aliasing of f1 as f2
    return f1(x, y);
}
```
or,

```cpp
#define f2 f1   // f2 will now be replaced by f1, effectively acting as an alias

int f1(int x, int y) {
    return x + y;
}
```

faheel (changes requested):

alxmjo:
@faheel Done.

faheel (approved):

PR #202, opened 2018-09-13
Migrate extended Euclidean algorithm

alxmjo:
Migrate extended Euclidean algorithm to header file (as described in #163).

- [x] Read the [contribution guidelines][contrib-guidelines]
- [x] Added [unit tests][unit-tests] (or unit tests have already been added)
- [x] Updated the [Contents][contents]

[contrib-guidelines]: https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md
[unit-tests]: https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md
[contents]: https://github.com/faheel/Algos/tree/master/C%2B%2B#contents

PR #224, opened 2018-10-19
Testing for GCD Algorithms

coopdog95:
I added testing files for gcd.cpp and extended_euclid.cpp, and prepped those files for the testing.


alxmjo:
Thanks for contributing, @coopdog95. We're actually contributing against the `reorganise` branch right now, which has a slightly different format. I also think we have tests written for those algorithms already (though they may be in open PRs).

One thing that we're trying to focus on right now is migrating from our old organization scheme to the new one. Have a look at #163 and see #202 for an example.  

Let me know if you have any questions! 🙂

PR #230, opened 2018-10-26
Euclidean Algorithm

apoorva-mk:
- [ ] Read the [contribution guidelines][contrib-guidelines]
- [ ] Added [unit tests][unit-tests] (or unit tests have already been added)
- [ ] Updated the [Contents][contents]

[contrib-guidelines]: https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md
[unit-tests]: https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md
[contents]: https://github.com/faheel/Algos/tree/master/C%2B%2B#contents

PR #232, opened 2018-10-26
Apoorva2

apoorva-mk:
I apologise for the minor error in my previous commit in the euclidean algorithm. I have now fixed and added the extended euclidean algorithm as well.


alxmjo:
Hi @apoorva-mk. Please see my comments under #230.

PR #234, opened 2018-10-30
Doubly linked list

korrat:
Added an implementation of a doubly linked list

- [x] Read the [contribution guidelines][contrib-guidelines]
- [x] Added [unit tests][unit-tests] (or unit tests have already been added)
- [x] Updated the [Contents][contents]

Closes #187

[contrib-guidelines]: https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md
[unit-tests]: https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md
[contents]: https://github.com/faheel/Algos/tree/master/C%2B%2B#contents

PR #235, opened 2018-11-05
doubly_linked_list.cpp

Rda99:
Doubly Linked List
-------------------
Doubly Linked List is a variation of the Linked list in which navigation is possible in both ways, either forward and backward easily as compared to Single Linked List.

### DIFF
--- a/C++/include/data_structures/doubly_linked_list.cpp
+++ b/C++/include/data_structures/doubly_linked_list.cpp
@@ -1,0 +1,301 @@
+/*
+ * C++ Program to Implement Doubly Linked List 
+ */
+#include<iostream>
+#include<cstdio>
+#include<cstdlib>
+/*
+ * Node Declaration
+ */
+using namespace std;
+struct node
+{
+    int info;
+    struct node *next;
+    struct node *prev;
+}*start;
+ 
+/*
+ Class Declaration 
+ */
+class double_llist
+{
+    public:
+        void create_list(int value);
+        void add_begin(int value);
+        void add_after(int value, int position);
+        void delete_element(int value);
+        void search_element(int value);
+        void display_dlist();
+        void count();
+        void reverse();
+        double_llist()
+        {
+            start = NULL;  
+        }
+};
+ 
+/*
+ * Main: Conatins Menu
+ */
+int main()
+{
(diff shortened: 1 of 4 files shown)


C++/include/data_structures/doubly_linked_list.cpp
```diff
@@ -0,0 +1,301 @@
...
+        }
+};
+ 
+/*
+ * Main: Conatins Menu
+ */
+int main()
+{
+    int choice, element, position;
+    double_llist dl;
```
faheel, line 44: The `main` function wouldn't be required as this would be a header file.

C++/include/data_structures/doubly_linked_list.cpp
```diff
@@ -0,0 +1,301 @@
...
+{
+    int info;
+    struct node *next;
+    struct node *prev;
+}*start;
+ 
+/*
+ Class Declaration 
+ */
+class double_llist
```
faheel, line 21: The class should be named `DoublyLinkedList`.

C++/include/data_structures/doubly_linked_list.cpp
```diff
@@ -0,0 +1,301 @@
...
+/*
+ * Node Declaration
+ */
+using namespace std;
+struct node
+{
+    int info;
+    struct node *next;
+    struct node *prev;
+}*start;
```
faheel, line 16: `*start` should be a private class variable that is initialized in its constructor.

faheel (changes requested):
@Rda99 Sorry for the late review. If you're still interested in contributing this implementation then here are a few changes I suggest you make:

* Convert `doubly_linked_list.cpp` into a header file (`doubly_linked_list.hpp`) and remove `main()` from it.
* Drop the commits from the `master` branch.
* Add unit tests in `C++/test/data_structures/doubly_linked_list.cpp` (see [UNIT_TEST.md](https://github.com/faheel/Algos/blob/reorganise/C%2B%2B/UNIT_TESTS.md) for more details).

More changes in the comments below.

alxmjo:
@Rda99 Thanks for your work on this, Raj. Closing since we merged doubly linked list in #234.

PR #242, opened 2019-06-02
Migrate Fibonacci

alxmjo:
- Add unit tests as cpp file
- Update readme
- Update build script
- Remove old files


faheel:
Nice work! Just wanted to point out that there's mixed usage of spaces and tabs here. But no need to worry about that now, we can fix that project-wide once all of the code is reorganised.

alxmjo:
Thanks for the close look. Maybe a carryover from my copying and pasting. I'll double-check that for future migrations.

PR #248, opened 2019-06-03
Migrate Sieve of Eratosthenes and add unit tests

alxmjo:
- Add unit tests as cpp file
- Update readme
- Update build script
- Remove old file


alxmjo:
@faheel I had to make a few decisions about how to migrate this over so I'll leave the PR open for a day in case you'd like to have a look before merge. Main change was returning primes in a vector rather than sending them to stdout.

PR #249, opened 2019-06-04
Migrate Bubble Sort and add unit tests

alxmjo:
- Add sorting cpp file (which can be reused for other sorting algos)
- Add include guard to utils.hpp
- Update readme
- Update build script
- Remove old file

@faheel This is my basic template for sorting algorithm migration, so I will leave this open for you until you have a look at it.


C++/test/algorithm/sorting/sorting.cpp
```diff
@@ -0,0 +1,82 @@
...
+	------------------------
+	Generates a vector with a random size (up to 100) and fills with random integers.
+*/
+vector<int> generate_unsorted_vector() {
+	vector<int> v;
+	auto vectorSize = (size_t) generate_random_int(1, 1000);
+	v.reserve(vectorSize);
+
+	for (int i = 0; i < (int) vectorSize; i++) {
+		v.push_back(generate_random_int(-32767, 32767));
```
faheel: I think we should increase the range to -2^31 to 2^31 - 1. Or it would be even better to use [`std::numeric_limits`](https://en.cppreference.com/w/cpp/types/numeric_limits).

C++/test/algorithm/sorting/sorting.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+/*
+	generate_random_int
+	-------------------
+	Generates a random int between min and max.
+*/
+int generate_random_int(int min, int max) {
+	std::default_random_engine generator(std::random_device{}());
+	std::uniform_int_distribution<> intRange(min, max);
+	return intRange(generator);
```
faheel: `intRange` should be renamed `int_range` to maintain naming consistency.

C++/test/algorithm/sorting/sorting.cpp
```diff
@@ -0,0 +1,82 @@
...
+			REQUIRE(algo_sorted == std_sorted);
+			algo_sorted = original;
+		}
+	}
+}
+
+/*
+	generate_unsorted_vector
+	------------------------
+	Generates a vector with a random size (up to 100) and fills with random integers.
```
faheel: Based on the code, the comment should say `up to 1000` (instead of 100). Also, I think the following will be better:

```
Creates a vector of random size (up to 1000) and populates it with random integers.
```

It would also be better if the function accepted an argument `max_size` and had 1000 as the default value for it if none was provided. This would make it more flexible, and therefore reusable in even more test case scenarios.

faheel (changes requested):

C++/test/algorithm/sorting/sorting.cpp
alxmjo: Noted. I'll use `std::numeric_limits<int>::min()`.

alxmjo:
Thanks for the careful review. I've implemented the suggested changes.

C++/test/algorithm/sorting/sorting.cpp
alxmjo: Was working on `counting_sort.hpp` today and realized that this wide range of integers is causing the algorithm to run extremely slowly ([more info](https://stackoverflow.com/questions/25001800/why-we-can-not-apply-counting-sort-to-general-arrays)). I'm going to use the following going forward:

`v.push_back(generate_random_int(std::numeric_limits<short>::min(), std::numeric_limits<short>::max()));`

This prevents the use of magic numbers while still giving a reasonably large range of positive and negative values to sort.

C++/test/algorithm/sorting/sorting.cpp
alxmjo: Forgot to add that if you have other suggestions I'm more than happy to follow suit.

C++/test/algorithm/sorting/sorting.cpp
faheel: Makes sense :+1:

PR #251, opened 2019-06-08
Migrate sorting algorithms

alxmjo:
I've moved over the rest of the sorting algorithms. The only anomaly (intentional anomaly, anyway) is that radix sort only sorts on positive integers, so I've left off the unit tests for now. I can add an issue to update the algorithm (using something [like this](https://stackoverflow.com/questions/15306665/radix-sort-for-negative-integers)).


C++/include/algorithm/sorting/radix_sort.hpp
```diff
@@ -0,0 +1,105 @@
...
+		display_state(values);
+	}
+}
+
+/*
+    max_in_vector
+    -------------
+    Returns the maximum value in a vector.
+*/
+int max_in_vect(const vector<int>& values) {
```
faheel, line 66: The function name should be `max_in_vector` (as in the comment).

C++/include/algorithm/sorting/radix_sort.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+	// On each iteration of the following loop, extractor helps in getting the
+	// next significant digit, which is (value / extractor) mod 10
+	for (int extractor = 1; max_value / extractor > 0; extractor *= 10) {
+		count_sort(values, extractor, to_show_state, mult_factor, add_factor);
+	}
+}
+
+// Wrapper function
+void radix_sort() {
```
faheel, line 95: Missing parameters in the function signature: `values`, `order`, and `to_show_state`.

C++/include/algorithm/sorting/shell_sort.hpp
```diff
@@ -0,0 +1,44 @@
...
+    O(1).
+*/
+
+#include <vector>
+#include "sorting/utils.hpp"
+
+using std::vector;
+
+void shell_sort(vector<int>& values, const int order = 1, const bool to_show_state = false) {
+	int n = values.size();
```
faheel, line 24: The following would be better:
```cpp
size_t size = values.size();
```

C++/include/algorithm/sorting/shell_sort.hpp
```diff
@@ -0,0 +1,44 @@
...
+	int temp;
+	int i, j;
+
+	// gap sequence is n/2, n/4, n/8, ..., 1
+	for (int gap = n / 2; gap > 0; gap /= 2) {
```
faheel, line 29: * Type of `i`, `j` and `gap` should be `size_t`.
* `i` and `j` can be declared in their respective for loop's init statement. It's good practice to limit the scope of variables. On the same note, `temp` can also be declared inside the i-loop's body.

C++/include/algorithm/sorting/merge_sort.hpp
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	size_t s = start;
+	for (int sorted_value: temp_vect)
+		values[s++] = sorted_value;
+}
+
+// Carry out merge sort
+void merge_sort_internal(vector<int>& values, const size_t start, const size_t end,
```
faheel, line 57: `merge_sort_internal` can be renamed to `merge_sort_range`, to signify that it's not just an internal function as it could be useful in scenarios where only a certain range of values in a vector need to be sorted.

faheel (changes requested):

faheel:
@alxmjo I had added some review comments yesterday but had forgotten to submit the review, and when I saw the open tab in my browser right now and finally submitted, only then did I see that the changes have already been merged.

I'll create a new issue to fix these changes later on.

alxmjo:
OK. I'm almost done migrating the remaining data structures. I can make these changes when I'm done with that.

C++/include/algorithm/sorting/radix_sort.hpp
alxmjo: Fixed in #253.

C++/include/algorithm/sorting/shell_sort.hpp
alxmjo: Fixed in #253.

C++/include/algorithm/sorting/shell_sort.hpp
alxmjo: Fixed in #253.

C++/include/algorithm/sorting/merge_sort.hpp
alxmjo: Fixed in #253.

alxmjo:
These fixes are implemented in #253.

PR #250, opened 2019-06-08
Added linked list algorithms

ghost:
Copy And Move Constructors have been added for SinglyLinkedList - following The Rule of Big Three, along with Copy and Swap for Assignment operator; some indentation issues have been resolved, also compilation errors have been removed, and Algorithm to reverse the list has been added.

- [x] Read the [contribution guidelines][contrib-guidelines]
- [ ] Added [unit tests][unit-tests] (or unit tests have already been added)
- [ ] Updated the [Contents][contents]

[contrib-guidelines]: https://github.com/faheel/Algos/blob/master/CONTRIBUTING.md
[unit-tests]: https://github.com/faheel/Algos/blob/master/C%2B%2B/UNIT_TESTS.md
[contents]: https://github.com/faheel/Algos/tree/master/C%2B%2B#contents

PR #252, opened 2019-06-12
Migrate data structures

alxmjo:
- Migrate singly linked list
- Migrate binary search tree

@faheel Besides the fixes you pointed out today and a few other little things here and there (migrating `utils.hpp` to the proper `sorting` folder, for one), I think this is the final step of converting all existing code into the new format. Only took us two years!  😄

I'll leave this open until you've had a chance to look it over.


alxmjo:
The clang++5.0 build failed with the following error code:

```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-5.0" failed and exited with 100 during .
```

Can't determine whether this is my problem or not. More info [here](https://travis-ci.community/t/build-fails-on-ubuntu-14-04-with-error-the-command-sudo-e-apt-get-yq-no-install-suggests-no-install-recommends-travis-apt-get-options-install-powershell-failed-and-exited-with-100-during/3116).

alxmjo:
I just re-ran the build associated with #250 without issue, which makes me think that there's something up with this PR. On the other hand, [this](https://travis-ci.community/t/cannot-apt-get-install-clang-5-0/3250) seems to indicate an issue with using Trusty as the default build environment. Apparently Travis is migrating to Xenial.

alxmjo:
The most recent commit appears to have done the trick. Specified Trusty manually, though it seems like at some point we'll want to [migrate this to Xenial](https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476).

.travis.yml
```diff
@@ -42,6 +42,7 @@ matrix:
     # Linux Clang 5 build
     - os: linux
       compiler: clang
+      dist: trusty
```
faheel: I think removing `dist` and `sources` should work as Clang 5.0 (along with 4.0 and 6.0) is available in the default Xenial repos (from the [link you shared](https://travis-ci.community/t/cannot-apt-get-install-clang-5-0/3250/3)).

For now, I suggest dropping the commits to this file and letting the build fail (since we know that the tests run fine otherwise) and then later updating the Travis CI config in a separate PR where we can also target GCC 8, Clang 6 and the latest macOS image.

C++/README.md
```diff
@@ -46,9 +46,9 @@ Implementation of well-known (and some rare) algorithms, in C++.
   * [Edit Distance Problem](include/algorithm/string/edit_distance.hpp) :white_check_mark:
 
 ### Data structures
-* [Binary search tree](include/data_structures/binary_search_tree.cpp)
-* [Disjoint-set](include/data_structures/disjoint_set.cpp) :white_check_mark:
-* [Singly linked list](include/data_structures/singly_linked_list.cpp)
+* [Binary search tree](include/data_structure/tree/binary_search_tree.hpp) :white_check_mark:
+* [Disjoint-set](include/data_structure/set/disjoint_set.hpp) :white_check_mark:
+* [Singly linked list](include/data_structure/linked_list/singly_linked_list.hpp) :white_check_mark:
```
faheel: Would be nice to arrange these under their respective topic's list, i.e. trees, sets and linked lists.

C++/include/data_structure/linked_list/singly_linked_list.hpp
```diff
@@ -0,0 +1,394 @@
...
+	Helper class for the nodes in the list
+*/
+
+template<class T>
+struct Node {
+	T value;
+	Node<T> *next;
+	Node(void);
+
+public:
```
faheel, line 29: This is not required as `struct` methods and members are public by default.

C++/include/data_structure/linked_list/singly_linked_list.hpp
```diff
@@ -0,0 +1,394 @@
...
+	Node(const T&, Node<T> * const);
+	Node(const Node&);
+	Node<T>& operator=(const Node<T> &);
+	~Node();
+	T& get_value(void);
+	void set_value(const T&);
+	Node<T> *get_next(void);
+	void set_next(Node<T> * const);
```
faheel: Getters and setters are usually not used for structures. Or is this supposed to be a class, in which case the `struct` declaration has been mistakenly used instead of `class`?

C++/include/data_structure/linked_list/singly_linked_list.hpp
```diff
@@ -0,0 +1,394 @@
...
+template<class T>
+struct Node {
+	T value;
+	Node<T> *next;
+	Node(void);
+
+public:
+	Node(const T&, Node<T> * const);
+	Node(const Node&);
+	Node<T>& operator=(const Node<T> &);
```
faheel: In the parameter type, `&` should not have a space before it.

faheel (changes requested):
At a few places, some function parameters and local variables are written as:
```cpp
Node<T> * const some_var
Node<T> & other_var
```
which are not wrong, but I think it would be better to write these as:
```cpp
const Node<T>* some_var
Node<T>& other_var
```
since its easier to determine the type of a variable (`const Node<T>*` and `Node<T>&`) this way.

C++/include/data_structure/linked_list/singly_linked_list.hpp
alxmjo: I think you're right that this was originally meant to be a class, so the `public` modifier makes more sense. With that in mind I've switched this to class and left in the modifier.

C++/include/data_structure/linked_list/singly_linked_list.hpp
alxmjo: Switched to class.

C++/README.md
alxmjo: Done.

.travis.yml
alxmjo: Reverted to the original format.

alxmjo:
Requested changes submitted, including the issues with the Node<T> references and pointers.

C++/test/data_structure/tree/binary_search_tree.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+TEST_CASE("Add and remove values", "[binary-search-tree]") {
+	BinarySearchTree tree;
+	tree.insert(1);
+	tree.insert(2);
+	REQUIRE(tree.search(1) == true);
+	tree.remove(1);
+	REQUIRE(tree.search(1) == false);
+	REQUIRE(tree.search(2) == true);
```
faheel: Having more insertions, removals and searches here would be good as that would make the test more robust, and therefore, more reliable.

C++/test/data_structure/linked_list/singly_linked_list.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+TEST_CASE("Create an empty linked list", "[singly-linked-list]") {
+	SinglyLinkedList<int> list;
+	REQUIRE(list.is_empty() == true);
+}
+
+TEST_CASE("Add elements", "[singly-linked-list]") {
+	SinglyLinkedList<int> list;
+	list.insert_rear(2);
+	list.insert_rear(3);
```
faheel: None of the tests cases singly linked list make use of `insert_front()`, `operator[]` or `is_empty()`. Calling these functions in various unit tests will improve code coverage.

faheel (changes requested):

C++/test/data_structure/tree/binary_search_tree.cpp
alxmjo: Good thing you asked for this, because it turns out that removing three elements in a row throws a SIGABRT. I'll run that down and add it to a new commit.

alxmjo: Looks like `remove()` was missing a return statement. Added a bunch of tests and they're all passing. Thanks again for the nudge.

alxmjo:
Changes and fixes implemented. Ready to merge as far as I'm concerned.

faheel (approved):

faheel:
Nice work @alxmjo!

PR #255, opened 2019-06-23
Reorganise C++ code into header files

### LINKED ISSUE #163: Reorganise C++ code into header files

faheel:
### Merge checklist

- [x] There are no links or references to the `reorganise` branch.
- [x] All tabs are replaced with 4 spaces in C++ code and 2 spaces and in Markdown.
- [x] All header files have include guards.
- [x] All unit tests pass.

closes #163.


alxmjo (commented):
Checked this out locally and ran all tests successfully. So on that front I think all of this looks good. 

The only things that stand out to me here are:

- Spaces (versus tabs). It sounded like you had a plan for doing this on a project-wide basis.
- I'm curious if it would make sense to merge #253 into `reorganise` before completing this merge into `master`. Maybe it doesn't make a difference.
- The `CONTRIBUTING.md` conflict referenced below.
- I also wonder if at some point it may be helpful to explicitly state that algorithms are implemented as header files under `/include` and that tests are added as .cpp files under `/test`. I know the `README` alludes to this but based on the number of PRs that don't follow this form I think being explicit would be good. 
- Getting off on a contributing tangent here, but I also think that it might be helpful if the `README` explicitly stated that contributors should start by opening an issue (as it says in the contribution guidelines).
- Maybe once we've merged into `master` I'll review the contribution guidelines and just open a PR with these changes implemented. 🙂

alxmjo:
I'm also happy to do the tab and spaces reformatting – just let me know. And if I do that, are there other formatting things we should standardize? The commenting is the only other thing I've noticed as being particularly inconsistent.

faheel:
@alxmjo I've replaced all tabs with spaces (except in the Makefile and in `catch.hpp`). It would be helpful if you could add include guards to some of the header files where they're missing.

Merging #253 into `reorganise` before merging this would be good. I'll get onto it.

Also, the changes you suggested in the CONTRIBUTING and README seem good, and as you suggested, we should implement them in a separate PR once this one gets merged.

alxmjo:
All header files now have include guards.

faheel:
It's finally done! :tada:

Thanks a lot for your contributions @alxmjo! This would've still been pending if it wasn't for you.

alxmjo:
Happy to help! 🙂

PR #260, opened 2019-07-28
Integrate Stale bot

faheel:


alxmjo (approved):
Looks good. Thanks for adding this.

alxmjo:
Having thought about this for ten more seconds, I wonder if another tag that we may also want to exempt are suggestions of new things to add. I think there's some utility to keeping open issues to give potential contributors ideas of what to add.

In a similar vein, I wonder if it would make sense to keep open issues with the tag `easy` or `up for grabs`.

faheel:
Good idea! I've exempted a few more labels.

alxmjo (approved):

PR #264, opened 2019-09-03
Overloading  ternary search to accept a function and an interval as parameters

### LINKED ISSUE #261: Accept a function as a parameter for ternary search
The way ternary search works now is that it takes in a unimodal array and the pattern that the elements in it follow (do they first increase then decrease or do they first decrease then increase) and then returns either the largest or the smallest value in the array (depending on the pattern).

A more general implementation for ternary would be like the one mentioned on [Wikipedia](https://en.wikipedia.org/wiki/Ternary_search#Algorithm) (and elsewhere too) that takes in the unimodal function `func` for which we want to find out the largest or smallest value in the given domain, along with the bounds of the domain, `start` and `end`.

It would be nice to have an overloaded definition for `ternary_search` that implemented this behaviour.

sylvaus:
Overloading ternary search to accept a function and an interval as parameters

Closes #261

### DIFF
--- a/C++/test/algorithm/searching/ternary_search.cpp
+++ b/C++/test/algorithm/searching/ternary_search.cpp
@@ -1,5 +1,4 @@
-#include <algorithm>
-#include <math.h>
+#include <cmath>
 
 #include "third_party/catch.hpp"
 #include "algorithm/searching/ternary_search.hpp"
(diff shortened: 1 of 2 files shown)


alxmjo:
Thanks for contributing @sylvaus. Sorry I haven't had a chance to review it yet but I will soon!

sylvaus:
My pleasure

C++/test/algorithm/searching/ternary_search.cpp
```diff
@@ -1,4 +1,5 @@
 #include <algorithm>
```
alxmjo: I don't think `algorithm` is necessary here.

C++/test/algorithm/searching/ternary_search.cpp
```diff
@@ -1,4 +1,5 @@
...
+#include <math.h>
```
alxmjo: Use `cmath` instead.

alxmjo (changes requested):
Thank you for this contribution. Sorry it took me so long to look it over.

In addition to the inline notes I've made, I think it would be valuable to add comments on the general ideas of what the code is doing, both in the implementation and the tests. You don't need to comment every single line, but have a look at some of the other algorithms for an idea for what to aim for.

Finally, I'll also note that there are some changes we could make in the unit test for ternary, but it doesn't make sense to hold up this PR for pre-existing issues so we identify those in a separate issue/PR.

sylvaus:
No problem, rework after review done

PR #266, opened 2019-09-25
Inserted a check for sorted input vector

pnjha:
If the input vector is already sorted we need not do O(N^2) computations for by simply putting an O(N) check.

### DIFF
--- a/C++/include/algorithm/sorting/bubble_sort.hpp
+++ b/C++/include/algorithm/sorting/bubble_sort.hpp
@@ -26,6 +26,20 @@ void bubble_sort(vector<int>& values, const int order = 1, const bool to_show_st
     bool swapped;
     size_t i, j;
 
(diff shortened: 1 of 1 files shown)


alxmjo:
Thanks for opening the PR, @pnjha, but I'm not sure that this is much of an improvement. 

The current implementation records whether elements have been swapped in the inner loop and records this in the variable `swapped`. If no elements have been swapped, then it breaks out of the sort early. In the event that an array is already sorted then the algorithm will complete after one pass through (which is essentially what your code does). 

Does this make sense? Or did I misunderstand the intent of your addition?

pnjha:
You are correct I missed that part. Your way of checking the array is already sorted or not is way better than what I proposed.

alxmjo:
@pnjha There are plenty of other places to add to the project, though, so please look around and don't be afraid to ask questions!

Issue #289, opened 2019-10-13
greatest_common_divisor_efficient


Paspasuy:
I want to add new function which will count GCD faster.

Paspasuy:
Not efficient function uses O(n) memory because it needs to save recursion stack, while this function uses O(1) memory.

PR #290, opened 2019-10-13
Added greatest_common_divisor_efficient algo

Paspasuy:
close issue #289 
Added greatest_common_divisor_efficient algo in algorithm/number_theory and unit tests.


alxmjo:
Thanks for your submission! I don't think "efficient" is the best word here, since efficient is a relative term. The existing GCD algorithm we have uses the Euclidean method, also, but is recursive. So maybe it makes sense to rename *that* algorithm `greatest_common_divisor_recursive` and then name this implementation `greatest_common_divisor`. Thoughts, @faheel?

PR #306, opened 2019-10-24
added bogo_sort

prasannathapa:


alxmjo:
Hi @prasannathapa. You're correct that there is a common sorting unit test, but if we tried to use it with bogo sort it might take a few million years before it completed. 🙂 

As such, I think it would be best to write a separate unit test for bogo sort which can actually be completed in a reasonable amount of time (a few seconds, for example).

PR #310, opened 2019-10-25
Update README.md

nomaan-2k:
Edited Readme.

Just added some punctuation marks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,15 +12,15 @@ Implementing well-known (and some rare) **algorithms and data structures**, whil
 
 ## How to get involved
 
-There are a few ways to get involved.
+There are a few ways to get involved:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,15 +12,15 @@ Implementing well-known (and some rare) **algorithms and data structures**, whil
 
 ## How to get involved
 
-There are a few ways to get involved.
```
alxmjo: Period is better here since a header follows immediately after the statement rather than an itemized list.

alxmjo (changes requested):
Fix the colon and it's good to go. 🙂

README.md
nomaan-2k: Sure

PR #324, opened 2019-11-18
Added Heap

### LINKED ISSUE #323: Data Structure: Heap
I implemented a heap tree based on the comparison function passed by a template parameter. Which will make a min heap or a max heap depending on this function. Also, added some test cases.

MarceloMoraesJr:
Added a heap tree data structure based on comparison function passed by template parameter. 

- [X] Read the [contribution guidelines][contrib-guidelines]
- [X] Added [unit tests][unit-tests] (or unit tests have already been added)
- [X] Updated the [Contents][contents]

 Closes #323

### DIFF
--- a/C++/CMakeLists.txt
+++ b/C++/CMakeLists.txt
@@ -178,3 +178,9 @@ target_link_libraries(disjoint_set test_runner)
 add_executable(binary_search_tree
         test/data_structure/tree/binary_search_tree.cpp)
 target_link_libraries(binary_search_tree test_runner)
+
+# Heap
+add_executable(heap
+        test/data_structure/heap/heap.cpp)
+target_link_libraries(heap test_runner)
+
--- a/C++/CMakeLists.txt
+++ b/C++/CMakeLists.txt
@@ -181,6 +181,6 @@ target_link_libraries(binary_search_tree test_runner)
 
 # Heap
 add_executable(heap
-        test/data_structure/heap/heap.cpp)
+        test/data_structure/tree/heap.cpp)
 target_link_libraries(heap test_runner)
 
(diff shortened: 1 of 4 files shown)


alxmjo:
I should be able to take a look at this today or tomorrow, but in the meantime, could you convert the tabs to spaces (one tab = 4 spaces)? Thanks!

alxmjo:
I'm still seeing plenty of tabs in your files, both in `heap.hpp` and `heap.cpp`. You should be able to use your IDE to convert these all at once. Also, in `heap.hpp`, lines 94 through 96, I also noticed a place where the double indentation was 7 spaces instead of 8. Please make sure these are fixed.

As for the tests in `heap.cpp`, have a look at how we implemented the tests in `binary_search_tree.cpp`. You'll want to make sure you use this same syntax (which draws on `Catch.hpp`), and that you're testing the equivalent functions in `heap.hpp` (insert, remove, etc.). 

It would also be a good idea to use the same naming conventions as `binary_search_tree.hpp`, like `insert()` instead of `Push_Heap()`, etc. And note the capitalization, too. Let's keep them all lower case.

Let me know if you have any questions!

alxmjo:
I see several people contributing to this PR. Out of curiosity, do you all know each other? Is this part of the class that Paolo was talking about?

MarceloMoraesJr:
Hi, i apologize for the delay in answering, it was a busy weekend. First, i would like to say thanks for your feedback on our PRs. Yes, we're part of Paulo's class, i'm sorry, i should have said that before. About the PR, we will try to fix it as soon as possible. Thank you for understanding!

alxmjo:
No problem, and no rush. 🙂

PR #333, opened 2020-05-20
Fix return type of linear search algo 

tryfinally:
linear-search will return index of item inside the vector, hence it must be of type size_t.
in case of not finding return an index that is one past the last element of the vector.
(same semantics as in C++ iterator end)

### DIFF
--- a/C++/include/algorithm/searching/linear_search.hpp
+++ b/C++/include/algorithm/searching/linear_search.hpp
@@ -22,16 +22,17 @@
     linear_search
     -------------
     Returns the index where a given element is found in an array. If the
-    element is not found, it returns -1.
+    element is not found, it returns index of element one past last one.
+    same as c++ iterator end style.
 */
 
 template <typename T>
-int linear_search(const T& element, const std::vector<T>& values) {
+size_t linear_search(const T& element, const std::vector<T>& values) {
     for (size_t i = 0; i < values.size(); i++)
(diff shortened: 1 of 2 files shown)


alxmjo:
Thanks for contributing, @tryfinally. I appreciate the thinking behind this, but I don't think it's actually an improvement for this particular project. 

I understand that your implementation is similar to what the STL does (though there's obviously a difference between returning the size of the container versus returning an iterator), but I think that it's actually less clear what's going on if you're unfamiliar with this search algorithm. For one, you need to know the size of the container before you can determine whether an item is present or not. With returning `-1`, on the other hand, it's immediately apparent.

In my experience, returning `-1` for not found (in a barebones implementation like this) is more common than returning the size of the array. It may not be the proper way of doing this according to the STL, but in this case I think it's more clear, especially for people who are new to the algorithm, new to testing, etc., which is who this project is oriented toward. There's also some good conversation on the subject [here](https://stackoverflow.com/questions/1970316/how-to-handle-item-not-found-situations-in-a-find-function).

tryfinally:
@alxmjo returning and int is jus wrong. On success you are returning index of item, which is size_t.
such an index might be unpresentable in int. The caller will get a negative value or worst some other
positive value which is not the index of the found element.
alternative, approach to correct this is by returning a pointer to the found element or null.

PR #337, opened 2020-05-28
Add Heap's algorithm

### LINKED ISSUE #336: Add Heap's algorithm for string permutations
I'd like to add the Heap's Algorithm to generate all permutations of a string. I've already implemented it according to the guidelines. Just need to polish some things up.

a-moreira:
I've added the Heap's Algorithm for generating all permutations of a string.

Closes #336 

### DIFF
--- a/C++/CMakeLists.txt
+++ b/C++/CMakeLists.txt
@@ -157,6 +157,11 @@ add_executable(shunting_yard
         test/algorithm/string/shunting_yard.cpp)
 target_link_libraries(shunting_yard test_runner)
 
+# Heap's Algorithm
+add_executable(heaps_algorithm
+        test/algorithm/string/heaps_algorithm.cpp)
+target_link_libraries(heaps_algorithm test_runner)
+
 # ============================================================================
 # Data structures
 # ============================================================================
--- a/C++/CMakeLists.txt
+++ b/C++/CMakeLists.txt
@@ -23,7 +23,7 @@ add_library(test_runner STATIC
 # Backtracking
 # -------------------
 
-# N-Queens
+# N-queens
 add_executable(n_queens
         test/algorithm/backtracking/n_queens.cpp)
 target_link_libraries(n_queens test_runner)
@@ -129,6 +129,7 @@ target_link_libraries(ternary_search test_runner)
 # Sorting
 # -------
 
+# Common sorting
 add_executable(sorting
         test/algorithm/sorting/sorting.cpp)
 target_link_libraries(sorting test_runner)
@@ -137,12 +138,12 @@ target_link_libraries(sorting test_runner)
 # String
 # ------
 
-# Edit distance Problem
+# Edit distance
 add_executable(edit_distance
         test/algorithm/string/edit_distance.cpp)
 target_link_libraries(edit_distance test_runner)
 
-# Heap's Algorithm
+# Heap's
 add_executable(heaps_algorithm
         test/algorithm/string/heaps_algorithm.cpp)
 target_link_libraries(heaps_algorithm test_runner)
--- a/C++/CMakeLists.txt
+++ b/C++/CMakeLists.txt
@@ -142,25 +142,25 @@ add_executable(edit_distance
         test/algorithm/string/edit_distance.cpp)
 target_link_libraries(edit_distance test_runner)
 
-# Longest common subsequence
-add_executable(longest_common_subsequence
-        test/algorithm/string/longest_common_subsequence.cpp)
-target_link_libraries(longest_common_subsequence test_runner)
+# Heap's Algorithm
+add_executable(heaps_algorithm
+        test/algorithm/string/heaps_algorithm.cpp)
+target_link_libraries(heaps_algorithm test_runner)
 
 # Knuth-Morris-Pratt
 add_executable(knuth_morris_pratt
         test/algorithm/string/knuth_morris_pratt.cpp)
 target_link_libraries(knuth_morris_pratt test_runner)
 
+# Longest common subsequence
+add_executable(longest_common_subsequence
+        test/algorithm/string/longest_common_subsequence.cpp)
+target_link_libraries(longest_common_subsequence test_runner)
+
 # Shunting yard
 add_executable(shunting_yard
         test/algorithm/string/shunting_yard.cpp)
 target_link_libraries(shunting_yard test_runner)
-
-# Heap's Algorithm
-add_executable(heaps_algorithm
-        test/algorithm/string/heaps_algorithm.cpp)
-target_link_libraries(heaps_algorithm test_runner)
 
 # ============================================================================
 # Data structures
--- a/C++/include/algorithm/string/heaps_algorithm.hpp
+++ b/C++/include/algorithm/string/heaps_algorithm.hpp
@@ -2,45 +2,57 @@
     Heap's Algorithm
     ----------------
 
-    This algorithm generates all possible
-permutations of a collection of n elements.    
-The algorithm minimizes movement: it generates each permutation from the previous one by interchanging a single pair of elements, while the other n-2 remain undisturbed.
+    This algorithm generates all possible permutations of a collection of n elements.    
+    The algorithm minimizes movement: it generates each permutation from the previous 
+    one by interchanging a single pair of elements, while the other n-2 remain undisturbed.
     
     Time Complexity
     ---------------
-    O(n!), where n is the number of elements in the collection
+    O(N!), where N is the number of elements in the collection
 
     Space Complexity
     ----------------
-    O(n), where n is the number of elements in the collection
+    O(N), where N is the number of elements in the collection
 
 */
-
 
 #ifndef HEAPS_ALGORITHM_HPP
 #define HEAPS_ALGORITHM_HPP
 
 #include <vector>
 
-using namespace std;
+using std::string;
+using std::vector;
+using std::swap;
 
 vector<string> heaps_algorithm(unsigned int number_of_elements, string& collection, vector<string>& permutations) {   
-
+    
     if (number_of_elements == 1 || collection == "") {
         permutations.push_back(collection);
     }
     
     else {
-        heaps_algorithm(number_of_elements-1, collection, permutations);
 
-        for (int i = 0; i < number_of_elements-1; i++) { 
-            if (number_of_elements % 2 == 0) {
-                swap(collection[i], collection[number_of_elements-1]);
+        /* Generate permutations with i-th element unaltered 
+         * Initially number_of_elements == length of input string */
+        heaps_algorithm(number_of_elements - 1, collection, permutations);
+
+        /* Generate permutations where the i-th element is swapped
+         * with each i-1 initial elements */
+        for (int i = 0; i < number_of_elements - 1; i++) { 
+            
+            /* The swap choice depends on the parity of 
+             * number_of_elements within each recursive call */
+            if (number_of_elements % 2 == 0) {        
+                swap(collection[i], collection[number_of_elements - 1]); // zero-indexed, the i-th is at i-1
+            
+            } else {
+                swap(collection[0], collection[number_of_elements - 1]);
             }
-            else {
-                swap(collection[0], collection[number_of_elements-1]);
-            }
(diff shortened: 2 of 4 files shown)


C++/CMakeLists.txt
```diff
@@ -157,6 +157,11 @@ add_executable(shunting_yard
         test/algorithm/string/shunting_yard.cpp)
 target_link_libraries(shunting_yard test_runner)
 
+# Heap's Algorithm
```
alxmjo: Just "Heap's". These are all algorithms so no need to point this out as such. Same goes for other naming, elsewhere. When naming files, though, just `heap` (so that we don't have to mess with apostrophes).

Also, note the alphabetical order.

C++/CMakeLists.txt
```diff
@@ -157,6 +157,11 @@ add_executable(shunting_yard
...
+add_executable(heaps_algorithm
```
alxmjo: Just `heap`

C++/CMakeLists.txt
```diff
@@ -157,6 +157,11 @@ add_executable(shunting_yard
...
+        test/algorithm/string/heaps_algorithm.cpp)
```
alxmjo: `heap.cpp`

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
+/*
+    Heap's Algorithm
```
alxmjo, line 2: This is okay as-is.

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+    ----------------
+
+    This algorithm generates all possible
+permutations of a collection of n elements.
```
alxmjo: Fix justification.

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+    Heap's Algorithm
+    ----------------
+
+    This algorithm generates all possible
+permutations of a collection of n elements.    
+The algorithm minimizes movement: it generates each permutation from the previous one by interchanging a single pair of elements, while the other n-2 remain undisturbed.
+    
+    Time Complexity
+    ---------------
+    O(n!), where n is the number of elements in the collection
```
alxmjo: capital n: `O(N!)`

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    Space Complexity
+    ----------------
+    O(n), where n is the number of elements in the collection
```
alxmjo: capital n

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+*/
+
+
+#ifndef HEAPS_ALGORITHM_HPP
+#define HEAPS_ALGORITHM_HPP
+
+#include <vector>
+
+using namespace std;
```
alxmjo: Instead of `using namespace std` (which pulls in entire `std` namespace), please pull in only the identifiers you need. i.e. 

```
using std::string;
using std::vector;
```

You can also just prepend the identifier with `std::` whenever you use it. As in, `std::swap(collection[0],`.

alxmjo: More info on this here: https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+        heaps_algorithm(number_of_elements-1, collection, permutations);
+
+        for (int i = 0; i < number_of_elements-1; i++) { 
+            if (number_of_elements % 2 == 0) {
+                swap(collection[i], collection[number_of_elements-1]);
+            }
+            else {
+                swap(collection[0], collection[number_of_elements-1]);
+            }
+            heaps_algorithm(number_of_elements-1, collection, permutations);
```
alxmjo: This is personal preference, but I find `variable - 1` to be more readable than `variable-1`.

C++/test/algorithm/string/heaps_algorithm.cpp
```diff
@@ -0,0 +1,41 @@
+#define CATCH_CONFIG_MAIN
+#include "third_party/catch.hpp"
+#include "algorithm/string/heaps_algorithm.hpp"
+
+
```
alxmjo, line 5: remove whitespace

C++/CMakeLists.txt
alxmjo: `# Heap's` (no need to identify it as an algorithm since that's implied by the category)

C++/CMakeLists.txt
alxmjo: At first I thought we might want to change this to just `heap` or `heaps`, but since that's also the name of a data structure, I think it's okay to leave this with the `_algorithm` suffix (for both function and file names) even though that's contrary to the naming conventions for the rest of the project.

C++/include/algorithm/string/heaps_algorithm.hpp
alxmjo: Fix justification

C++/include/algorithm/string/heaps_algorithm.hpp
alxmjo: `N` instead of `n`

C++/include/algorithm/string/heaps_algorithm.hpp
alxmjo: `N` instead of `n`

C++/include/algorithm/string/heaps_algorithm.hpp
alxmjo: Instead of `using namespace std` (which brings in the entire std namespace), just bring in what you need, i.e.:

```
using std::vector;
using std::string;
```

You can also just prepend `std::` whenever you use those identifiers, as in: `std::swap(collection[0],...`

alxmjo: More info here: https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

C++/include/algorithm/string/heaps_algorithm.hpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+    
+    else {
+        heaps_algorithm(number_of_elements-1, collection, permutations);
+
+        for (int i = 0; i < number_of_elements-1; i++) { 
+            if (number_of_elements % 2 == 0) {
+                swap(collection[i], collection[number_of_elements-1]);
+            }
+            else {
```
alxmjo: `} else {`

C++/include/algorithm/string/heaps_algorithm.hpp
alxmjo: Personal preference, but I think `var - 1` is more readable than `var-1`.

C++/test/algorithm/string/heaps_algorithm.cpp
```diff
@@ -0,0 +1,41 @@
...
+TEST_CASE("Base cases", "[string][heaps_algorithm]") {
+    vector<string> output_1;
+    string s = ""; // empty string
```
alxmjo, line 8: Again, personal preference, but I'd prefer `str` over `s`

C++/test/algorithm/string/heaps_algorithm.cpp
```diff
@@ -0,0 +1,41 @@
...
+    REQUIRE(heaps_algorithm(s.length(), s, output_2) == permutations);
+    
+    vector<string> output_3;
+    s = " ";
+    permutations = {" "};
+    REQUIRE(heaps_algorithm(s.length(), s, output_3) == permutations);
+}
+
+
+TEST_CASE("short cases", "[string][heaps_algorithm]") {
```
alxmjo, line 24: might be nice to have another short case that includes a space character

C++/test/algorithm/string/heaps_algorithm.cpp
```diff
@@ -0,0 +1,41 @@
...
+    vector<string> permutations = {"abc", "bac", "cab", "acb", "bca", "cba"};
+    REQUIRE(heaps_algorithm(s.length(), s, output_4) == permutations);
+
+    vector<string> output_5;
+    s = "a$1";
+    permutations = {"a$1", "$a1", "1a$", "a1$", "$1a", "1$a"};
+    REQUIRE(heaps_algorithm(s.length(), s, output_5) == permutations);
+}
+
+TEST_CASE("long cases", "[string][heaps_algorithm]") {
```
alxmjo, line 36: I know that it's pretty onerous to write out these longer cases, but it would be nice to have another one or two.

alxmjo (changes requested):
This looks good. Lots of great code and organization, etc. A few minor requests and suggestions included, plus two overarching requests: 

1. It would be nice to have a few explanatory comments in your `.hpp` file.
2. It would be better to organize your PRs so that they aren't overlapping. Since you updated the readme in your other PR, this PR is technically incomplete. If we merged the readme PR first (which I'll review after I finish this), then it would reference an algorithm that doesn't actually exist yet. And obviously if we merge this first, then the readme wouldn't be accurate. So if it's not too much trouble, use this branch/PR to add the algorithm to the readme, and use your other branch/PR to update the other parts of the readme. Hopefully that makes sense.

Let me know if you have questions, and thanks!

C++/CMakeLists.txt
a-moreira: I agree, both because of ambiguity and because the algorithm is indeed named after B. R. Heap.

C++/CMakeLists.txt
a-moreira: I'll fix this and also exchange LCS and KMP places for the alphabetical order to be correct.

C++/test/algorithm/string/heaps_algorithm.cpp
a-moreira: I prefer `s` over `str` because `str()` is a standard library function. See https://www.cplusplus.com/reference/sstream/stringstream/str/

C++/include/algorithm/string/heaps_algorithm.hpp
a-moreira: Thanks! I chose `using std::` for the sake of clean code.

C++/test/algorithm/string/heaps_algorithm.cpp
alxmjo: Of course. Thanks. 🙂

C++/test/algorithm/string/heaps_algorithm.cpp
a-moreira: OK

a-moreira:
Added the tag "Permutation" since there are other (interesting) algorithms to generate them.

a-moreira:
> Added the tag "Permutation" since there are other (interesting) algorithms to generate them.

Maybe I should've opened an issue to discuss that. What do you think?

alxmjo:
Pushed a couple extra commits to fix some things that I didn't notice the first time around. Also used this merging opportunity to clean up `CMakeLists.txt` a bit. Will merge once checks pass. Thank you!

C++/test/algorithm/string/heaps_algorithm.cpp
alxmjo: "Algos" was a nice touch here. 🙂

alxmjo:
Also just wanted to say that this was a great submission. I hope you find other things to add! 🙂

a-moreira:
@alxmjo Just found a little mistake in the last test that went unnoticed by both of us, there's no REQUIRE() function. Already fixed it locally, does it need another pull request or can this one be reopened?

alxmjo:
Good catch. Probably easiest to just open a new PR. I'll merge tonight.

PR #341, opened 2020-05-29
Add data conversion functions

### LINKED ISSUE #99: Implement data validation
Could we add some data validation functions such as, 
make upper case,
make lower case,
get random double,
get random int?

beardbytes:
Test cases for data validation function

#99 


C++/include/utils/data_validation.hpp
```diff
@@ -0,0 +1,64 @@
...
+    ------------
+    string having all the characters in upper case .
+*/
+
+string make_upper_case(string str)
+{
+    //converting each character to upper case using for loop
+    for (auto &c : str)
+    {
+        c = std::toupper(c); // using the function toupper() for conversion
```
alxmjo: Rather than just using `std::toupper()` and `std::tolower()`, I think it would be better to implement this manually (as done [here](https://github.com/mrKappen/Algos/blob/662de4b9a254fe5da228c93334549b7fcf9e2f7d/C%2B%2B/include/utils/data_validation.hpp), though there are lots of ways to do it). Obviously in "real life" you'd just use the `std` function, but the point of this repo is (generally) to implement things ourselves (i.e., sorting a list manually rather than just calling `std::sort()`). More good info [here](https://stackoverflow.com/questions/22425825/changing-a-lowercase-character-to-uppercase-in-c).

So, please implement these utilities without using the `std` functions. Let me know if you have any questions!

alxmjo (changes requested):

C++/include/utils/data_validation.hpp
beardbytes: Okay...I will dive right into it !

alxmjo (changes requested):
This is a good start! A few things to comment on. 

I'm subscribed to the PR automatically, so no need to tag me in a separate comment saying it's ready for review.

These are actually conversion functions, so replace file names and comments with `data_conversion`. 

While you've written the unit tests, you need to add them to `CMakeLists.txt` in order to run them. Add the following to the bottom of that file:

```
# ============================================================================
# Utilities
# ============================================================================

# Data conversion
add_executable(data_conversion
        test/utils/data_conversion.cpp)
target_link_libraries(data_conversion test_runner)
```

This will show you that some of your tests aren't actually passing.

Finally, I think that rather than using test cases to break out the tests into base and non-base cases, you should use the test cases for each function, like this:

```
TEST_CASE("Make upper case", "[string][data-validation][make-upper-case]")
{
    // upper case base cases

    // upper case other cases
}

TEST_CASE("Make lower case", "[string][data-validation][make-lower-case]")
{
    // lower case base cases

    // lower case other cases
}

```

Finally, a base case should test empty strings and strings with one character in them. Everything else can go under the other comment. And make sure to test other edge cases. Like, what happens with numbers (ints, floats)? What happens with special characters? What happens with spaces?

Let me know if you have any questions.

beardbytes:
> This is a good start! A few things to comment on.
> 
> I'm subscribed to the PR automatically, so no need to tag me in a separate comment saying it's ready for review.
> 
> These are actually conversion functions, so replace file names and comments with `data_conversion`.
> 
> While you've written the unit tests, you need to add them to `CMakeLists.txt` in order to run them. Add the following to the bottom of that file:
> 
> ```
> # ============================================================================
> # Utilities
> # ============================================================================
> 
> # Data conversion
> add_executable(data_conversion
>         test/utils/data_conversion.cpp)
> target_link_libraries(data_conversion test_runner)
> ```
> 
> This will show you that some of your tests aren't actually passing.
> 
> Finally, I think that rather than using test cases to break out the tests into base and non-base cases, you should use the test cases for each function, like this:
> 
> ```
> TEST_CASE("Make upper case", "[string][data-validation][make-upper-case]")
> {
>     // upper case base cases
> 
>     // upper case other cases
> }
> 
> TEST_CASE("Make lower case", "[string][data-validation][make-lower-case]")
> {
>     // lower case base cases
> 
>     // lower case other cases
> }
> ```
> 
> Finally, a base case should test empty strings and strings with one character in them. Everything else can go under the other comment. And make sure to test other edge cases. Like, what happens with numbers (ints, floats)? What happens with special characters? What happens with spaces?
> 
> Let me know if you have any questions.

Alright . I will keep that in mind . 

>  These are actually conversion functions, so replace file names and comments with `data_conversion` ->> Which file names to replace ?

alxmjo:
- [ ] Replace `data_validation.hpp` with `data_conversion.hpp`
- [ ] Replace `data_validation.cpp` with `data_conversion.cpp`
- [ ] Fix any comment that references data validation.

beardbytes:
Alright ! Got it

beardbytes:
> See the red circle with the line through it on the line above? Make sure to add a newline at the end of all of your files. See here: https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

Okay ! Got it .

alxmjo:
See in this thread under "Changes requested" where it says "All checks have failed?" At least one of your unit tests is currently failing, so please fix that. 

Also, remove the old `data_validation` files. 

Finally, please break out the unit tests into base cases (with empty string and single character) and other cases, with everything else. These don't need to be entirely other test cases, just a comment separating, like so:

```
TEST_CASE("Make upper case", "[string][data-validation][make-upper-case]")
{
    // Upper case base cases
    REQUIRE(...)
    REQUIRE(...)
    
    // Upper case other cases
    REQUIRE(...)
    REQUIRE(...)
    ...
}
```

beardbytes:
> See in this thread under "Changes requested" where it says "All checks have failed?" At least one of your unit tests is currently failing, so please fix that.
> 
> Also, remove the old `data_validation` files.
> 
> Finally, please break out the unit tests into base cases (with empty string and single character) and other cases, with everything else. These don't need to be entirely other test cases, just a comment separating, like so:
> 
> ```
> TEST_CASE("Make upper case", "[string][data-validation][make-upper-case]")
> {
>     // Upper case base cases
>     REQUIRE(...)
>     REQUIRE(...)
>     
>     // Upper case other cases
>     REQUIRE(...)
>     REQUIRE(...)
>     ...
> }
> ```

Yes...I am figuring out why the tests are failing . 
Okay i will separate the cases from base cases !

beardbytes:
**I am getting this error**
`g++ data_conversion.cpp
data_conversion.cpp:1:10: fatal error: third_party/catch.hpp: No such file or directory
 #include "third_party/catch.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.`

alxmjo:
You still have `data_validation` files included in your commit. Remove those. As for your error, I checkouted out all of your files on my machine and I cannot reproduce it. Make sure you're in `C++` directory and then run `make` and then `make test`. You should see that one of your tests is still failing.

PR #345, opened 2020-06-02
Update compilation directions and information

### LINKED ISSUE #344: Command make all does not work
[README.md](https://github.com/ProAlgos/ProAlgos-Cpp/blob/42060df5388e99bd8ab3dfaaa4afef67358fe288/C%2B%2B/README.md) in the `C++` directory gives the following directions for building: 

> To compile the source files, run **`make all`**.

However, when I download a fresh copy of the repo and run the command, I see the following:

```
alex ProAlgos-Cpp-master $ cd C++
alex C++ $ make all
make: *** No rule to make target `all'.  Stop.
```

Running `make` and `make default` both appear to have the desired effect. So, I think we should either update the `Makefile` or update `README.md` so that the directions are accurate.

Tagging @faheel since he originally set up CMake on the project.

alxmjo:
Update documentation to match current make configuration. Closes #344.

### DIFF
--- a/C++/README.md
+++ b/C++/README.md
@@ -76,8 +76,8 @@ Implementing well-known (and some rare) **algorithms and data structures in C++*
 
 ## Compiling
 
-To compile the source files, run **`make`** from the `C++` directory. Doing so will create executable binaries in 
-sub-directories under the `bin` directory.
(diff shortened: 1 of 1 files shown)


alxmjo:
Closing as these changes have been integrated into #347.

PR #349, opened 2020-06-09
Add primorial algorithm

### LINKED ISSUE #43: Primorial
Implement an algorithm for calculating the [primorial](https://en.wikipedia.org/wiki/Primorial) of a given prime number _P<sub>n</sub>_

nseabourn:

Closes #43

### DIFF
--- a/C++/README.md
+++ b/C++/README.md
@@ -34,6 +34,7 @@ Implementing well-known (and some rare) **algorithms and data structures in C++*
   * [Perfect number check](include/algorithm/number_theory/perfect_number_check.hpp) :white_check_mark:
   * Prime numbers
     * [Sieve of Eratosthenes (simple)](include/algorithm/number_theory/sieve_of_eratosthenes.hpp) :white_check_mark:
+    * [Primorial](include/algorithm/number_theory/primorial.hpp) :white_check_mark:
   
 * Searching
   * [Binary search](include/algorithm/searching/binary_search.hpp) :white_check_mark:
--- a/C++/README.md
+++ b/C++/README.md
@@ -33,8 +33,8 @@ Implementing well-known (and some rare) **algorithms and data structures in C++*
     * [Logarithmic time algorithm](include/algorithm/number_theory/fibonacci_efficient.hpp) :white_check_mark:
   * [Perfect number check](include/algorithm/number_theory/perfect_number_check.hpp) :white_check_mark:
   * Prime numbers
-    * [Sieve of Eratosthenes (simple)](include/algorithm/number_theory/sieve_of_eratosthenes.hpp) :white_check_mark:
     * [Primorial](include/algorithm/number_theory/primorial.hpp) :white_check_mark:
+    * [Sieve of Eratosthenes (simple)](include/algorithm/number_theory/sieve_of_eratosthenes.hpp) :white_check_mark:
   
 * Searching
   * [Binary search](include/algorithm/searching/binary_search.hpp) :white_check_mark:
--- a/C++/include/algorithm/number_theory/primorial.hpp
+++ b/C++/include/algorithm/number_theory/primorial.hpp
@@ -1,16 +1,20 @@
 /*
     Primorial
     ---------
-    Calculate the primorial Pn# of the nth prime number Pn.
+    Calculate the primorial Pn# for the prime number definition of primorial 
+    and n# for the natural number definition of primorial.
     Pn# is defined as the product of the first n primes.
+    n# is defined as the product of all primes less than or equal to n.
 
     Time complexity
     ---------------
-    O(Pn^2) where Pn is the nth prime number.
+    O(N^2) where N is n for the natural definition and Pn for the prime 
+    definition.
 
     Space complexity
     ----------------
-    O(N).
+    O(N) where N is n for the natural definition and Pn for the prime 
+    definition.
 */
 
 #ifndef PRIMORIAL_HPP
--- a/C++/include/algorithm/number_theory/primorial.hpp
+++ b/C++/include/algorithm/number_theory/primorial.hpp
@@ -21,16 +21,18 @@
 
 typedef unsigned long long int ULL;
 
-// primorial_Pn(Pn) where Pn > 47 goes beyond the range of ULL
-const unsigned int MAX_P_N = 47;
+// primorial_natural(n) where n > 52 goes beyond the range of ULL
+const unsigned int MAX_N_NATURAL = 52;
 
-// primorial_n(n) where n > 15 goes beyond the range of ULL
+// primorial(n) where n > 15 goes beyond the range of ULL
 const unsigned int MAX_N = 15;
 
-// returns 0 if Pn is not a prime number or if Pn# is too big to fit in an 
-// ULL. returns the primorial Pn# otherwise
-ULL primorial_Pn(unsigned int p_n) {
-    if (p_n < 2) {
+// computes the primorial defined by natural numbers: https://oeis.org/A034386
+// returns 0 if n# is too big to fit in an ULL.
+// returns the primorial n# otherwise
+ULL primorial_natural(unsigned int n) {
+    // check for unsigned integer wraparound
+    if (n > MAX_N_NATURAL) {
         return 0;
     }
 
@@ -38,7 +40,7 @@ ULL primorial_Pn(unsigned int p_n) {
     
     ULL product = 1;
 
-    for (unsigned int i = 2; i <= p_n; i++) {
+    for (unsigned int i = 2; i <= n; i++) {
         bool is_prime = true;
         for (std::vector<unsigned int>::iterator it = primes.begin(); is_prime
             && it != primes.end(); it++) {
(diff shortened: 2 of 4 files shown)


C++/README.md
```diff
@@ -34,6 +34,7 @@ Implementing well-known (and some rare) **algorithms and data structures in C++*
   * [Perfect number check](include/algorithm/number_theory/perfect_number_check.hpp) :white_check_mark:
   * Prime numbers
     * [Sieve of Eratosthenes (simple)](include/algorithm/number_theory/sieve_of_eratosthenes.hpp) :white_check_mark:
+    * [Primorial](include/algorithm/number_theory/primorial.hpp) :white_check_mark:
```
alxmjo, line 36: Alphabetize

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
+/*
+    Primorial
+    -----------------
```
alxmjo: Match hyphens to title (should be 9, I think).

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+#define PRIMORIAL_HPP
+
+#include <climits>
+#include <vector>
+
+typedef unsigned long long int ULL;
+
+const unsigned int MAX_P_N = 47;   // primorial(Pn) where Pn > 47 goes beyond the range of ULL
+
+//returns 0 if p_n is not a prime number or if p_n_# is too big to fit in an ULL. returns the primorial p_n_# otherwise
```
alxmjo: Space between `//` and start of comment, and please put comment on two lines.

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+ULL primorial(unsigned int p_n){
+    if(p_n < 2)
+        return 0;
```
alxmjo, line 40: Use curly braces.

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    std::vector<unsigned int> primes;
+    
+    ULL product = 1;
+
+    for(unsigned int i=2; i<=p_n; i++){
```
alxmjo: For all `if`, `for`, `while`, etc.:

```
if () {
    // code
}
``` 

Note the curly braces and space between `if` and condition and between condition and first brace.

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+        bool is_prime = true;
+        for(std::vector<unsigned int>::iterator it = primes.begin(); is_prime && it != primes.end(); it++){
```
alxmjo: See above notes on formatting.

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+            if(i % *it == 0)
+                is_prime = false;
+        }
+        if(is_prime){            
+            //check for unsigned integer wraparound
+            if(ULLONG_MAX / product < i)
```
alxmjo:
```
if () {
    // code
} else {
    // code
}
```

C++/include/algorithm/number_theory/primorial.hpp
```diff
@@ -0,0 +1,59 @@
...
+            else{
+                product *= i;
+                primes.push_back(i);
+            }
+        }
+    }
+
+    //make sure Pn was a prime nubmer
+    if(primes[primes.size()-1] != p_n)
+        return 0;
```
alxmjo, line 69: Add curly braces

C++/test/algorithm/number_theory/primorial.cpp
```diff
@@ -0,0 +1,31 @@
...
+    REQUIRE(primorial(2) == 2);
+}
+
+TEST_CASE("Normal cases", "[primorial]") {
+    REQUIRE(primorial(3) == 6);
+    REQUIRE(primorial(5) == 30);
+    REQUIRE(primorial(7) == 210);
+    REQUIRE(primorial(11) == 2310);
+    REQUIRE(primorial(13) == 30030);
+    REQUIRE(primorial(MAX_P_N) == 614889782588491410);
```
alxmjo: I wonder if it would be worth declaring in a comment that `MAX_P_N` is equal to 47 (since it's defined in the header)?

C++/test/algorithm/number_theory/primorial.cpp
```diff
@@ -0,0 +1,31 @@
+#include "third_party/catch.hpp"
+#include "algorithm/number_theory/primorial.hpp"
+
+TEST_CASE("Base cases", "[primorial]") {
+    REQUIRE(primorial(2) == 2);
```
alxmjo: Shouldn't this be equal to `6`?

alxmjo: Since we want the first two primes and `2 * 3 = 6`.

alxmjo (changes requested):
Thanks for contributing, @nseabourn! 

It looks like you're actually working off of two distinct definitions of primorial. In the comments in `primorial.hpp` you use the definition that Pn# is the "product of the first n primes." But then in the tests we have cases like `primorial(3) = 6` when, based on your definition, it would be `primorial(3) = 2 * 3 * 5 = 30`. The [Wikipedia page](https://en.wikipedia.org/wiki/Primorial) lists two definitions, and I think we should go with the first (which matches your comment in `primorial.hpp`). Also see the [Wolfram page](https://mathworld.wolfram.com/Primorial.html) and the [OEIS page](https://oeis.org/wiki/Primorial#Primorial_of_natural_numbers) (I think this is the clearest explanation on the difference between the two definitions). 

That said, I'm not an authority on this be any means so I'm happy to hear your thoughts.

alxmjo:
Also, out of curiosity, are you a student at Oregon State? I noticed your CS 362 repo and it took me back. 😁

alxmjo:
Just realized that there is some commentary on method of implementation in #43. Still, I think the most straightforward and consistent way to implement this is to take `n` as input and then return the product of the first `n` primes.

alxmjo:
There are several ways one could handle taking `Pn` as input, too. Maybe split into two functions: `primorial_n()` taking `n` as input and also `primorial_Pn()` taking `Pn` as input? 

As I said, I'm not an authority, but I do want to make sure that we are clear and consistent with whatever method we end up with.

nseabourn:
Thank you for all of your input.
For your question about the test cases, the parameter of this function is Pn, the nth prime number. So primorial(2) is just 2, and primorial(11) = 2*3*5*7*11 = 2310.
To alleviate the confusion I can do what you suggested and make two different functions, primorial_n(n) and primorial_p_n(p_n). So primorial_p_n would be what I already wrote and primorial_n(2) = 2*3 = 6.
I will also fix what I wrote so that it adheres to the code style and add a comment for MAX_P_N .
And yeah, I am a student at Oregon State, I'm just about to graduate actually. Did you go there as well?

alxmjo:
Let's go with `primorial_Pn()` (or possibly `primorial_pn()` if you think it'd be better to keep it lower case). I think the second underscore makes it less readable.

I graduated from Oregon State last year. 🙂

alxmjo:
I like how you did this. A couple things:

**First:** I think that some of your values are off. Like, `primorial_n(0)` should be equal to 1, by definition. Likewise, `primorial_Pn(0) = primorial_Pn(1) = 1` by definition. The rest of the `primorial_n()` values look correct. 

As for `primorial_Pn()`, if the input is nonprime, we should just be calculating the product of the primes _less than_ that number. So, `primorial_Pn(6) = 30` because the primes less than or equal to 6 are 2, 3, and 5, and 2 * 3 * 5 = 30. Likewise, `primorial_Pn(7) = 7 * 5 * 3 * 2 = 210`.

The Wikipedia page for Primorial has a helpful table [here](https://en.wikipedia.org/wiki/Primorial#Table_of_primorials). Both `primorial_n()` and `primorial_Pn()` should take the **n** (first column) as input. Output of `primorial_n()` should be column **Pn#** and output of `primorial_Pn()` should be column **n#**. I think I have that right.

**Second:** I think it would be ideal to reorganize the unit tests. `Base cases` should include n = 0, 1, 2. Then `normal cases`. Then, I don't know, `edge cases` for the maximum computable values and maximum values + 1? Sort of like you have with wraparound cases, but based on what I wrote above they shouldn't really "wrap around." If `n` is too large for what the program can handle then I think it makes sense to return 0, as you've done. Again, based on what I wrote above, the only time you would receive 0 as output is as an exception because n is too large. I see you've already noted this in a comment for the function, which is great.

Finally, I think it would also be valuable to group the tests by function and separate them with a comment, as in:

```
TEST_CASE(…) {
    // primorial_n
    REQUIRE(…)
    REQUIRE(…)

    // primorial_Pn
    REQUIRE(…)
    REQUIRE(…)
}
```

I write too much. 😛

nseabourn:
Thanks for noticing that primorial_n(0) is wrong. I will fix that. I will also reorganize the test cases.
As for primorial_Pn, I think we are looking at two different definitions. I was using the definition for prime numbers: [https://en.wikipedia.org/wiki/Primorial#Definition_for_prime_numbers](https://en.wikipedia.org/wiki/Primorial#Definition_for_prime_numbers) whereas I think you are using the definition for natural numbers: [https://en.wikipedia.org/wiki/Primorial#Definition_for_natural_numbers](https://en.wikipedia.org/wiki/Primorial#Definition_for_natural_numbers).
These calculate Pn# and n# respectively. For Pn#, Pn must be a prime number. n# doesn't have that requirement. If n = Pn, then n# = Pn#. But if n is not a prime number then n# = Pn# where Pn is the closest prime number less than n. I hope that was easy to understand, I don't like how n was used in both definitions in different ways.
I can easily change primorial_Pn to fit the definition for n# since I would only need to change the output when the input isn't a prime number. I would like to change the function names primorial_Pn to primorial_n# and primorial_n to primorial_Pn# though and have both functions take the input n.

alxmjo:
You're right. I was misunderstanding your intent. I do think it makes sense to have one function that follows the `n`th prime number definition (input: `n`, output: `Pn#`) and then another function that follows the natural number definition (input: `n`, output: `n#`).

I think my original suggestion of using `…n` and `…Pn` led to our misunderstanding, so maybe we should avoid that naming scheme. We can't use a `#` sign in a function definition, though. 

Why not `primorial()` for the first definition (since I think that this is the most common way of defining primorial – see the [Rosetta Code page](https://rosettacode.org/wiki/Primorial_numbers) for primorial) and `primorial_natural()` for the second definition? 

It might also be nice to link to the OEIS pages for each series in the comments of each function: [A002110](https://oeis.org/A002110) for the first definition and [A034386](https://oeis.org/A034386) for the second.

nseabourn:
Good idea. I will do that.

alxmjo:
This is a great PR, @nseabourn. I really appreciate the addition and the conversation. Made one final commit on some things I forgot to mention the first time around. Will merge as soon as it passes CI checks.

alxmjo (approved):

PR #355, opened 2020-06-16
Add graph data structure using adjacency matrix

### LINKED ISSUE #335: Implement data structures for graph algorithms
In order to add graph algorithms, we must first add graph data structures. We'll need data structures for both directed and undirected graphs. We'll want to be able to add vertices and edges. In the end it'll be valuable to have implementations for both an adjacency list and an adjacency matrix.

If you're interested in adding this please explain your plan below so that we can get started. 🙂  

See #70 for previous conversation on the subject.

a-moreira:

Before proceeding to write the unit tests and the adjacency lists, I'd like some thoughts on the adjacency matrix. 
For example, whether there's a more efficient way to write it, whether the STL containers used are the appropriate ones, _if_ and _what_ should these class methods return.
Also, does this class need a method to remove vertices? (may be too expensive, so if the user needs a lot of removals, maybe the adjacency list should be used) etc

[unit-tests]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/.github/CONTRIBUTING.md#adding-unit-tests
[contents]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/.github/README.md#contents

### DIFF
--- a/cpp/include/data_structure/graph/adjacency_matrix.hpp
+++ b/cpp/include/data_structure/graph/adjacency_matrix.hpp
@@ -1,0 +1,175 @@
+#ifndef ADJACENCY_MATRIX_HPP
+#define ADJACENCY_MATRIX_HPP
+
+#include <iostream>
+#include <vector>
+#include <map>
+
+using std::vector;
+using std::map;
+
+template<typename Vertex>
+class Graph {
+    
+    public:
+        
+        Graph(bool is_directed);
+        virtual ~Graph()       ;
+        
+        /* set methods */
+        
+        // returns the index of the added vertex
+        // or -1 if it already exists
+        int add_vertex(const Vertex& v);
+
+        // returns EXIT_SUCCESS
+        // or -1 if it already exists
+        int add_edge   (int weight, const Vertex& v1, const Vertex& v2);
+        int remove_edge(const Vertex& v1, const Vertex& v2)            ;
+   
+
+
+        /* get methods */  
+        int get_index               (const Vertex& v)                   ;
+        int get_weight              (const Vertex& v1, const Vertex& v2);
+        int get_num_vertices        ()                                  ;
+        vector<Vertex>& get_vertices()                                  ; 
+    
+    private:
+
+        // this integer identifies each vertex in our map.
+        // we increase it every time we add a vertex. 
+        // It never decreases
+        int                  index   ;
+        vector<vector<int> > matrix  ;
+        map<int, Vertex>     indices ;
+        bool                 directed;
+};
+
+template<typename Vertex>
+Graph<Vertex>::Graph(bool is_directed) {
+    index = -1;
+    directed = is_directed;
+}
+
+template<typename Vertex>
+Graph<Vertex>::~Graph() {
+    indices.clear();
+    matrix.clear();
+    index = -1;
+}
+
+
+template<typename Vertex>
+int Graph<Vertex>::get_index(const Vertex& v) {
+    for (auto it : indices) {
+        if (it.second == v) {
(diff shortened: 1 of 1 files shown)


alxmjo (commented):
I think that this is a great start. Here are a few over-arching comments:

- You cover the basic functionality of the graph well. As for your question about being able to remove vertices, that's always something that can be added later on by you or another contributor.
- Please follow the code formatting laid out in [singly-linked list](https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/cpp/include/data_structure/linked_list/singly_linked_list.hpp).
- I like the comments you have so far. For functions, though, I think it's better to comment on functionality where the function is defined rather then where it's declared. Also the block comment above `index` could be reformatted as an inline comment so that it doesn't appear to apply to all member variables.
- This implementation assumes a weighted graph. One could create an unweighted graph using the same weight for all edges, but I wonder if it would be better to either rewrite `add_edge()` so that `weight` is an optional argument, or to pass in a bool when creating a graph object declaring whether the graph is weighted or unweighted (and then using that value appropriately). This second option is a bit more explicit, though would require writing more code. Having thought about it for a few seconds I'm indifferent, I'd just say that if you go with the former I think it'd be worth being explicit in the comments regarding how one would use this for an unweighted graph.
- Regarding your comments on if and what to return, I think your implementation thus far makes sense. Let me know if you have more specific questions. As for returning values for non-existent indices, some of our searching functions which return an index will return `-1` for an item not being found. I'm open to your thoughts on this since I don't think we've settled on a good solution that we can implement consistently in our project.
- Just a personal preference, but I like `index` as a variable name over `idx`, and `vertex_1` over `v1`. Since we're oriented toward the beginner end of the spectrum I'd gladly sacrifice brevity for readability.

Sorry if I've missed anything. Let me know as you have more questions. Looking forward to seeing the tests.

a-moreira:
@alxmjo Thank you very much for the suggestions.

- Regarding the weighted/unweighted matter: I think it's best if we go with weighted as default, since every unweighted graph simply has weight 1 for all edges. This way we'll keep it broader, tidier (less code), but also we'll add helpful formality (I guess anyone studying graphs should understand this idea). I'll explain this both in the comments and in the docs.

- About the return value for missing items, I think `-1` is the most straightforward way to do this and I don't think we are losing readability here. It could be the case that returning a string as an error serves better the idea of clarity first, but in the end I think standardization is much more helpful to achieve this goal. IMO the best way is to explain everything clearly and add all the details in the documentation, leaving the code as clean as possible. The only problem I can think of is if the function returns integers and thus possibly -1. Anyway this case should be an exception and we can solve the ambiguity in the docs.

alxmjo:
> @alxmjo Thank you very much for the suggestions.
> 
> * Regarding the weighted/unweighted matter: I think it's best if we go with weighted as default, since every unweighted graph simply has weight 1 for all edges. This way we'll keep it broader, tidier (less code), but also we'll add helpful formality (I guess anyone studying graphs should understand this idea). I'll explain this both in the comments and in the docs.
> * About the return value for missing items, I think `-1` is the most straightforward way to do this and I don't think we are losing readability here. It could be the case that returning a string as an error serves better the idea of clarity first, but in the end I think standardization is much more helpful to achieve this goal. IMO the best way is to explain everything clearly and add all the details in the documentation, leaving the code as clean as possible. The only problem I can think of is if the function returns integers and thus possibly -1. Anyway this case should be an exception and we can solve the ambiguity in the docs.

I appreciate your thoughts on both of these issues. Sounds like you have what you need to move forward. Let me know if you have other thoughts or questions.

Issue #376, opened 2020-10-01
Printing all the permutations of a string/array.


brijeshsos66:
In the Backtracking section, I would like to add CPP code for Printing all the permutations of a string/array. The Worst-case Time complexity is O(N*N!) & Space Complexity is O(N!).

brijeshsos66:
I tried it got failed, I will try again to make it work.

PR #377, opened 2020-10-01
Added backtracking algorithm named all permutations of a string/array

brijeshsos66:
Added **all_permutations.hpp** in include folder & **all_permutations.cpp** in test folder.

Issue #376

### DIFF
--- a/cpp/include/algorithm/backtracking/all_permutations.hpp
+++ b/cpp/include/algorithm/backtracking/all_permutations.hpp
@@ -1,0 +1,51 @@
+/*
+    Printing all the permutations of a string/array
+    --------------------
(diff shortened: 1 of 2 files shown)


alxmjo:
This looks good on first glance, but please read the [documentation regarding testing code](https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/docs/CONTRIBUTING.md#testing-code) (you need to update `CMakeLists.txt` so that we can test your code). Also, there is an existing README for backtracking algorithms (in the same directory where you added the algorithm implementation). Please update that.

alxmjo:
Sorry for the late response, @brijeshsos66, but I just realized that you made those changes in another PR that has since been closed. Can you add those changes on this branch so they're contained in the same PR?

PR #394, opened 2020-11-20
Added String Hash Function Implementation

### LINKED ISSUE #391: String Hashing Function
I would like to add a string hashing algorithm to allow for string comparison in O(1) time.

KyleSmith19091:
I've implemented a polynomial rolling hash function that can be applied to strings, given a string of any length the function will return a hash, which can be used for comparing strings, used in the Robin-Karp pattern matching algorithm or calculating the number of palindromic substrings in a string.

Closes #391 

[contrib-guidelines]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/docs/CONTRIBUTING.md#contribution-guidelines
[unit-tests]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/docs/CONTRIBUTING.md#testing-code
[contents]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/README.md#contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ handle it from there. :smile:
   * [Shunting yard](cpp/include/algorithm/string/shunting_yard.hpp) :white_check_mark:
   * Permutation
     * [Heap's Algorithm](cpp/include/algorithm/string/heaps_algorithm.hpp) :white_check_mark:
-  * [Rolling Polynomial Hash Function](cpp/include/algorithm/string/string_hash.hpp) :white_check_mark:  
+  * [Polynomial Rolling Hash Function](cpp/include/algorithm/string/string_hash.hpp) :white_check_mark:  
 
 ### Data structures
 
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ handle it from there. :smile:
   * [Shunting yard](cpp/include/algorithm/string/shunting_yard.hpp) :white_check_mark:
   * Permutation
     * [Heap's Algorithm](cpp/include/algorithm/string/heaps_algorithm.hpp) :white_check_mark:
+  * [Rolling Polynomial Hash Function](cpp/include/algorithm/string/string_hash.hpp) :white_check_mark:  
 
 ### Data structures
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ handle it from there. :smile:
   * [Shunting yard](cpp/include/algorithm/string/shunting_yard.hpp) :white_check_mark:
   * Permutation
     * [Heap's Algorithm](cpp/include/algorithm/string/heaps_algorithm.hpp) :white_check_mark:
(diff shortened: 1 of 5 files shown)


cpp/include/algorithm/string/string_hash.hpp
```diff
@@ -0,0 +1,46 @@
...
+// hash(s) = s[0] + s[1]*p + s[1]*p^2 ... mod m
+
+const int PRIME = 53; // Constant value used in polynomial rolling hash function 
+const long long MOD = 1e9 + 9; // Arbitrary large prime number to use as modulo, used to avoid collisions
+
+int compute_hash(std::string const& s){
+    long long hash_value = 0; // Return value
+    long long prime_power = 1; 
+
+    if(s.find_first_not_of(' ')==std::string::npos){ // Consider strings containing spaces as having hash value of 0
```
alxmjo: I'd move this comment up onto its own line.

cpp/include/algorithm/string/string_hash.hpp
```diff
@@ -0,0 +1,46 @@
+/*
+    String Hashing Function
+    ----------------------------------------------
```
alxmjo: Adjust hyphens so they match widths of section titles.

README.md
```diff
@@ -93,6 +93,7 @@ handle it from there. :smile:
   * [Shunting yard](cpp/include/algorithm/string/shunting_yard.hpp) :white_check_mark:
   * Permutation
     * [Heap's Algorithm](cpp/include/algorithm/string/heaps_algorithm.hpp) :white_check_mark:
+  * [Polynomial Rolling Hash Function](cpp/include/algorithm/string/string_hash.hpp) :white_check_mark:
```
alxmjo: My instinct is to change this to "String hashing" as I don't think this level of specificity (nor the implementation details) are necessary here. Thoughts?

alxmjo (changes requested):
This looks good. I was surprised to realize that we didn't actually have a hashing algorithm, so this is a nice addition. A few things I'd like you to change in order to stay in line with our practices:

**Imports**
I don't think the `iostream` include is necessary.

**Spacing in control blocks** 
```
if (this) {
    // then that
} else {
    // this other thing
}
``` 

```
for (char c : s) {
...
```
 
**Spacing in comparisons**
`s.find_first_not_of(' ') == std::string::npos`

**Only one newline at end of files**
In the testing file and implementation files, specifically.

**Match filenames between implementation and testing file**
One is `string_hash` the other is `string_hashing`.

**`compute_hash()` return type**
It's supposed to return an `int` but `hash_value` is a `long long`.

There's also at least one inline comment. Apologies if I've forgotten something.

KyleSmith19091:
Apologies for these silly mistakes I will correct them accordingly and also thank you for the detailed explanation!

README.md
KyleSmith19091: I agree with this statement, the specific implementation is not important as a hashing algorithm just returns some form of a hash so I will change the name to String Hashing.

PR #396, opened 2020-12-14
Data conversion

### LINKED ISSUE #99: Implement data validation

0xwal:
Implementation of data conversion from upper case to lower case and vice versa using
bits manipulation and wrote tests for all possible cases.
Both functions takes const ref to avoid copying and construct a local string variable
then return it with implicit move semantic without changing the input. 
I did all my best to comply with the guidelines.

I am open to any change requests on this PR and please detail the changes with reasons so I can learn. 🥰

Thanks.

Closes #99

[contrib-guidelines]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/docs/CONTRIBUTING.md#contribution-guidelines
[unit-tests]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/docs/CONTRIBUTING.md#testing-code
[contents]: https://github.com/ProAlgos/ProAlgos-Cpp/blob/master/README.md#contents

### DIFF
--- a/cpp/include/utils/data_conversion.hpp
+++ b/cpp/include/utils/data_conversion.hpp
@@ -1,0 +1,58 @@
+/*
+    Data Conversion
+    -------------------
+    Upper/Lower case algorithm using bit manipulation.
+
+    Time complexity
+    ---------------
+    O(N), where N is the length of the input.
+
+    Space complexity
+    ----------------
+    O(N), where N is the result.
+ */
+
+#ifndef DATA_CONVERSION_HPP
+#define DATA_CONVERSION_HPP
+
+#include <string>
+
+#define UPPER_CASE_BIT_MASK 0b11011111
+#define LOWER_CASE_BIT_MASK 0b00100000
+
+bool is_letter(const char& c) {
+    char lower_cased_char = (char)(c | LOWER_CASE_BIT_MASK);
+    return lower_cased_char >= 'a' && lower_cased_char <= 'z';
+}
+
+std::string make_upper_case(const std::string& input) {
+    std::string result;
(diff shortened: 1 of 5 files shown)


cpp/include/utils/data_conversion.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef DATA_CONVERSION_HPP
+#define DATA_CONVERSION_HPP
+
+#include <string>
+
+#define UPPER_CASE_BIT_MASK 0b11011111
+#define LOWER_CASE_BIT_MASK 0b00100000
+
+bool is_letter(const char& c) {
```
0xwal: char& by ref 😄, auto generated by Jetbrain Clion because the `&` being used in `for` loop line 30 & 45

faheel: Well spotted! All of these `char&` variables can be converted to `char` because creating references creates pointers, which would add unnecessary overhead for `char` values.

0xwal: Removed the ref from all char and all tests passes in my machine.

0xwal:
@faheel @alxmjo sorry for the mention, it took me some time to do this just for contribution and I would love to hear a feedback from you. I know you are busy and excuse me for this. 😊

0xwal:
@faheel  Sorry is supposed to be discarded ?

0xwal:
Not worth my time. should be archived if not maintained. Thanks @faheel , @alxmjo

PR #407, opened 2021-01-26
Update Catch to v2.13.4

faheel:
The Catch library hadn't been updated for almost 3 years so I've updated it to the latest v2.13.4.

### DIFF
--- a/cpp/include/third_party/catch.hpp
+++ b/cpp/include/third_party/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.1.2
- *  Generated: 2018-02-09 17:05:21.506253
+ *  Catch v2.13.4
+ *  Generated: 2020-12-29 14:48:00.116107
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2018 Two Blue Cubes Ltd. All rights reserved.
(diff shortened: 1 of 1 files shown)


alxmjo:
Thanks for this. It passes compilation and tests on GitHub, but when I run it locally I get the following:

```
...
weighted_activity_selection:
===============================================================================
All tests passed (9 assertions in 2 test cases)


real    0m0.005s
user    0m0.002s
sys     0m0.001s
make: *** [test] Error 1
```
I also get several instances of this:

```
cut_the_rod:
scripts/run_tests.sh: line 8: bin/cut_the_rod: Permission denied
```

To be clear, this doesn't appear to be an issue with this PR or the updated Catch framework since I get the same errors on `master`, but I'd like to have it sorted before committing. Any idea what's going on?

alxmjo:
Now that I think of it, the tests showing permission denied are not actually included in the repo. They're old PRs that were never actually merged. Thinking this is a CLion issue or some sort of artifact with how I have it set up.

alxmjo (approved):

alxmjo:
I have my own issues to sort out (looks to be related to CMake cache) but that appears to be unrelated to this PR. Thanks for updating. I'll leave it to you to merge.

faheel:
Cool, I'll merge this then. Good luck with solving your build issue! I've started work on #408 to test on multiple combinations of OSes, compilers and C++ versions so that we can be sure that the builds work as expected on different configurations.

PR #450, opened 2023-04-06
binary_search fix

pm100:
simple fix for https://github.com/ProAlgos/ProAlgos-Cpp/issues/449

if we were down to a slice with size == 1 then code tried to access slice[1]


faheel:
The following seems like a simpler fix as it prevents the function call for the second time if the value was not found, and also does not require checking the size in each recursive call:

```diff
- else if (low <= high)
+ else if (low < high)
```

pm100:
quite right :-)

PR #451, opened 2023-06-27
fix windows build and add sanitizers and tests

pm100:
This change does 3 things

- it makes the test build work on windows
- it adds ctest tests so that the bash script is not needed to run the tests (doesnt work on windows)
- it sets the sanitizer flags on the gcc build (since these are test / demo the overhead doesnt matter)


pm100:
need to fix it a bit more

